PR #57, opened 2015-04-15
Exposed classes under the GitGraph namespace.

ddidier:
The following classes are now exposed:
- GitGraph.Branch
- GitGraph.Commit
- GitGraph.Template

This enables custom enhancements.


nicoespeon:
Sounds fair. That way devs could add custom modifications easily if they feel to.

PR #68, opened 2015-06-16
Fix "Cannot set property offsetX of MouseEvent..." for non-Firefox browsers

timlegrand:
Fix the "Cannot set property offsetX of #<MouseEvent> which has only a getter" error for non-Firefox browsers.

Check that current browser is Firefox was taken from http://stackoverflow.com/questions/9847580/how-to-detect-safari-chrome-ie-firefox-and-opera-browser

### DIFF
--- a/src/gitgraph.js
+++ b/src/gitgraph.js
@@ -305,11 +305,13 @@
     var isOut = true;
 
     // Fix firefox MouseEvent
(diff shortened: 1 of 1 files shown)


timlegrand:
Hi Nico, sorry for the mess, I accidentally closed this pull request. Feel free to reopen it if you want. #69  includes it anyway.

PR #69, opened 2015-06-16
Add branch name at beginning of commit messages

timlegrand:
Add branch name at beginning of commit messages (by default, so `metro` template shows it in example file).
Can be explicitly enabled/disabled through the `templateConfig.commit.message.displayBranch` option.

Assume that [Pull Request 68](https://github.com/nicoespeon/gitgraph.js/pull/68) has been previously applied (Firefox fix). Not a requirement though.

### DIFF
--- a/examples/index.js
+++ b/examples/index.js
@@ -15,6 +15,7 @@ var myTemplateConfig = {
     },
     message: {
       displayAuthor: true,
+      displayBranch: false,
       displayHash: false,
       font: "normal 12pt Arial"
     }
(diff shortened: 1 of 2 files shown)


nicoespeon:
Hi @timlegrand !

Thanks so much for your help on this. I agree adding the name of the branch by default is a good idea. I'll just add a little word about it (and how to disable it) in the docs.

Thanks for the FF fix too, I'll check that and include it through this PR, no need to re-open the other one ;-)

PR #72, opened 2015-07-15
Fix the column calculation algorithm

sourtin:
After the first merge, the original algorithm will assign every new branch the same column, resulting in collisions. Fix to handle arbitrary numbers and orders of merges and branches.


nicoespeon:
Hi @sourtin !

Thanks for this one. Works nicely and make me found another bug I'll fix right away.
Thanks again :+1:

PR #76, opened 2015-07-28
Added the ability to assign 'onClick' events to commits. 

terhechte:
This required a bit of refactoring in order to utilize the already-existing hover functionality which serves to fetch the currently hovered commit.
I also added a jsdoc `commitCallback` type that is used twice. Once as a parameter for the new `applyCommits` function which serves as an abstracted way of determining the commit related to the current mouse event, and it also serves as the `onClick` parameter for the Commit. 

Due to the refactoring, the actual implementation of the onClick event is rather simple:

``` Javascript
GitGraph.prototype.click = function ( event ) {
    this.gitgraph.applyCommits(event, function(commit, isOverCommit) {
      if (isOverCommit && !(commit.onClick === null)) {
        commit.onClick(commit, true);
      }
    });
  };
```

So it would be trivially easy to also add support for right click, and others. I may go on and implement additional event handlers, but first I wanted to make sure that the way I structured this is ok with you and you'd be willing to merge this.

Cheers & Thanks for a great library!

### DIFF
--- a/src/gitgraph.js
+++ b/src/gitgraph.js
@@ -275,6 +275,8 @@
       unscaledResolution.x += 800;
     }
 
+    unscaledResolution.x += this.template.commit.widthExtension;
+
     this.canvas.style.width = unscaledResolution.x + "px";
     this.canvas.style.height = unscaledResolution.y + "px";
 
@@ -998,6 +1000,7 @@
    * @param {Number} [options.branch.spacingY] - Space between branchs
    * @param {Number} [options.commit.spacingX] - Space between commits
    * @param {Number} [options.commit.spacingY] - Space between commits
+   * @param {Number} [options.commit.widthExtension = 0]  - Additional width to be added to the calculated width
    * @param {String} [options.commit.color] - Master commit color (dot & message)
    * @param {String} [options.commit.dot.color] - Commit dot color
    * @param {Number} [options.commit.dot.size] - Commit dot size
@@ -1053,6 +1056,7 @@
     this.commit = {};
     this.commit.spacingX = options.commit.spacingX || 0;
     this.commit.spacingY = (typeof options.commit.spacingY === "number") ? options.commit.spacingY : 25;
+    this.commit.widthExtension = (typeof options.commit.widthExtension === "number") ? options.commit.widthExtension : 0;
 
     // Only one color, if null message takes branch color (full commit)
     this.commit.color = options.commit.color || null;
--- a/src/gitgraph.js
+++ b/src/gitgraph.js
@@ -332,11 +332,12 @@
    **/
   GitGraph.prototype.applyCommits = function(event, callbackFn) {
     // Fix firefox MouseEvent
-    event.offsetX = event.offsetX ? event.offsetX : event.layerX;
-    event.offsetY = event.offsetY ? event.offsetY : event.layerY;
-    event.x = event.x ? event.x : event.clientX;
-    event.y = event.y ? event.y : event.clientY;
-
+    if ( typeof InstallTrigger !== "undefined" )/* == (is Firefox) */ {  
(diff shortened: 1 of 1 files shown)


nicoespeon:
Hi @terhechte !

Thanks a lot for this. I'll take a closer look in the next days, but sounds an looks good so far :+1:

terhechte:
Awesome :)

src/gitgraph.js
```diff
@@ -307,6 +315,40 @@
...
+   * Hover event on commit dot
+   *
+   * @param {MouseEvent} event - Mouse event
+   * @param {commitCallback} callbackFn - A callback function that will be called for each commit
+   *
+   * @self Gitgraph
+   **/
+  GitGraph.prototype.applyCommits = function(event, callbackFn) {
+    // Fix firefox MouseEvent
+    event.offsetX = event.offsetX ? event.offsetX : event.layerX;
```
nicoespeon: @terhechte > refactoring looks fine, but what happens to the Firefox `if` guard? :smiley: 

```javascript
if ( typeof InstallTrigger !== "undefined" )/* == (is Firefox) */ {       		
  event.offsetX = event.offsetX ? event.offsetX : event.layerX;		
  event.offsetY = event.offsetY ? event.offsetY : event.layerY;		
  event.x = event.x ? event.x : event.clientX;		
  event.y = event.y ? event.y : event.clientY;		
}
```

src/gitgraph.js
```diff
@@ -317,22 +359,18 @@
...
-      event.y = event.y ? event.y : event.clientY;
-    }
-
-    function showCommitTooltip () {
+    function showCommitTooltip (commit) {
       self.tooltip.style.left = event.x + "px"; // TODO Scroll bug
       self.tooltip.style.top = event.y + "px";  // TODO Scroll bug
-      self.tooltip.textContent = commit.sha1 + " - " + commit.message;
+      if (!(self.template.commit.tooltipHTMLFormatter === null)) {
+        self.tooltip.innerHTML = self.template.commit.tooltipHTMLFormatter(commit);
```
nicoespeon: @terhechte > I see you also added the ability to format tooltips. That's nice, I'll check and document that too :+1:

src/gitgraph.js
```diff
@@ -363,13 +396,32 @@
...
+
+  /**
+   * Click event on commit dot
+   *
+   * @param {MouseEvent} event - Mouse event
+   *
+   * @self Gitgraph
+   **/
+  GitGraph.prototype.click = function ( event ) {
+    var self = this.gitgraph;
```
nicoespeon: @terhechte > I'm not sure that's still relevant since there is only one use of `self` here, at the same scope level.

src/gitgraph.js
```diff
@@ -363,13 +396,32 @@
...
+    self.applyCommits(event, function(commit, isOverCommit) {
+      if (isOverCommit) {
```
nicoespeon: @terhechte > I tend to prefer writing 

```
if(!isOverCommit) return;

if (!(commit.onClick === null)) {
  commit.onClick(commit, true);
}
```

I found that's a bit more readable (less nesting).

src/gitgraph.js
terhechte: Yeah, I think I just forgot that. I've just removed it.

terhechte:
I just pushed a new commit that includes all your proposes changes. 

Cheers!
Benedikt

nicoespeon:
Hi there!

I started looking at this PR before going on holiday in August, but didn't have time to complete. 
September was a bit busy for me too.

So I just want to say that I didn't forget you and will look at it back as soon as I would have some free time aside =)

Thanks for the patience :+1:

nicoespeon:
So far I've tested it works well.
Thanks for the patience, I've been very busy this end of year :+1:

PR #77, opened 2015-08-01
Fix scaling in inverse orientation

ringlej:
In an inverse orientation, if the scalingFactor is anything but 1, then parts of the graph will get cut off.
This patch fixes this by moving the context.scale call to after doing the inverse orientation translation.
Also, applies the scalingFactor to the margins

Signed-off-by: Jon Ringle <jringle@gridpoint.com>


nicoespeon:
Will be merged with #79

PR #78, opened 2015-08-02
Improve tag and message alignment

ringlej:
The tag text width calculation before calling drawTextBG was calculating
a different width than the drawTextBG, because different context.font was
set at the time.

Also keep track of the maxTagWidth and use it to determine where the
message text should go.

Signed-off-by: Jon Ringle <jringle@gridpoint.com>


nicoespeon:
Will be merged with #79

Issue #86, opened 2015-12-03
Branch from a previous commit


Martin1994:
These codes won't work properly:
 ````javascript
var graph = new GitGraph();
var master = graph.branch("master");
var branch = master.branch("branch");
master.commit();
branch.commit();
````

Should looks like this:
````
o--o
 \----o
````

But looks like this:
````
o--o
    \-o
````

orao:
I would like to provide an additional use case for this issue. _Though not standard git flow I do have a need for this funky merge type in my [ac2git](https://github.com/orao/ac2git) conversion._ The following is an [example of the incorrect rendering](http://jsfiddle.net/guan1oz1/468/), while I would expect it to look something like [this workaround](http://jsfiddle.net/guan1oz1/470/).

Additionally, there is one more problem that occurs when I want to swap the order of the merge. Once again, the [incorrect rendering example](http://jsfiddle.net/guan1oz1/472/) and the [workaround](http://jsfiddle.net/guan1oz1/473/).

If I get time I will take a look at these specific problems and provide a patch but it probably won't happen this week.

PR #88, opened 2015-12-03
Branch from current commit or a specific commit

Martin1994:
#86 

Changes:
* Add `parentCommit` option parameter to `Branch`;
* Default the parent commit to the latest commit of the parent branch before creating this branch (instead of the latest one before the first commit of this branch);
* Add `pushPath` to push points into `path` in order to avoid duplicate points (which causes unexpected short line at the duplicate point in bezier mode).

Test code:
````javascript
<html>
<head>
<script type="text/javascript" src="js/gitgraph.js"></script>
<script type="text/javascript">

function test() {
    var canvas = document.getElementById('canvas');
    
    var template = new GitGraph.Template({
        commit: {
            spacingY: -40
        }
    });
    
    var graph = new GitGraph({
        canvas: canvas,
        orientation: 'vertical',
        template: template
    });
    
    var master = graph.branch('master');
    master.commit();
    branch = master.branch('branch');
    master.commit();
    branch.commit();
    var branch2 = master.branch('branch2');
    branch2.commit();
    master.commit();
    branch2.branch('branch3').commit();
    master.merge(branch);
    
    var master2 = graph.orphanBranch('new master');
    master2.commit();
    master2.branch('new branch').commit();
    master2.commit();
}

</script>
</head>
<body onload="test()">
<canvas id="canvas" width="640" height="480">
</canvas>
</body>
</html>
````

Current develop:
screen shot 2015-12-03 at 2 40 49 pm

This PR:
screen shot 2015-12-03 at 2 40 26 pm

### DIFF
--- a/src/gitgraph.js
+++ b/src/gitgraph.js
@@ -477,7 +477,8 @@
    *
    * @param {Object} options - Options of branch
    * @param {GitGraph} options.parent - GitGraph constructor
-   * @param {Branch} [options.parentBranch] - Parent branch
+   * @param {Branch} [options.parentBranch = options.parentCommit.branch] - Parent branch
+   * @param {Commit} [options.parentCommit = options.parentBranch.commits.slice(-1)[0]] - Parent commit
    * @param {String} [options.name = "no-name"] - Branch name
    *
    * @this Branch
@@ -491,7 +492,22 @@
     // Options
     options = (typeof options === "object") ? options : {};
     this.parent = options.parent;
-    this.parentBranch = options.parentBranch;
+    if ( options.parentCommit && options.parentBranch ) {
+      if ( options.parentCommit.branch !== options.parentBranch ) {
+        return;
+      }
+      this.parentCommit = options.parentCommit;
+      this.parentBranch = options.parentBranch;
+    } else if ( options.parentCommit ) {
+      this.parentCommit = options.parentCommit;
+      this.parentBranch = options.parentCommit.branch;
+    } else if ( options.parentBranch ) {
+      this.parentCommit = options.parentBranch.commits.slice( -1 )[ 0 ];
+      this.parentBranch = options.parentBranch;
+    } else {
+      this.parentCommit = null;
+      this.parentBranch = null;
+    }
     this.name = (typeof options.name === "string") ? options.name : "no-name";
     this.context = this.parent.context;
     this.template = this.parent.template;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by nicoespeon** [comment]
Hi @Martin1994 and thanks a lot for this.

Just to tell you I will review all this within the next days as I get some free time. Will let you know!

**Comment by nicoespeon** [comment]
Hi @Martin1994, I did review the PR and it resolve few display quirks passing by, which is great :ok_hand: 

However I've got some regression issues with the [gitflow example](https://github.com/nicoespeon/gitgraph.js/blob/develop/examples/gitflowsupport.js) which doesn't play nice with your modifications.

The expected output, current state on `develop`:

image

The output from this PR:

image

Please let me know if you fix the issue / if you need help on this one.
Again, thanks for this :+1:

File: src/gitgraph.js [hunk 1]
```diff
@@ -776,6 +816,37 @@
...
+   * @this Branch
+   */
+  Branch.prototype.pushPath = function (point) {
+    var lastPoint = this.path.slice( -1 )[ 0 ];
+    if ( !lastPoint ) {
+      this.path.push( point );
+    } else if ( lastPoint.x === point.x && lastPoint.y === point.y ) {
+      if ( lastPoint.type !== "start" && point.type === "end" ) {
+        lastPoint.type = "end";
+      } else if ( point.type === "join" ) {
```
nicoespeon: I guess this `else if` / `else` could be refactor into one `else if ( point.type !== "join" )`

Martin1994:
Hi @nicoespeon, Thanks for the review. I will update this PR as soon as I'm free.

bpatra:
Excellent work, I have incorporated this commit on my fork https://github.com/bpatra/gitgraph.js/commit/db14c8043b1d8172114aba5bb965276215f5fe2f
I looks to work well and its mandatory (to me at least) because otherwise the gitgraph are not correct and cannot be displayed.

sdefauw:
:+1:

larshp:
Hi,
Is it possible to merge this fix? I have the same problem

nicoespeon:
@larshp > yup, you did well re-activating this thread. I got some free time, I’ll have a new look at it.

nicoespeon:
Hmm there still is the regression I mentioned earlier.

@Martin1994 did you have any time to have a look at it yet? The thing is that it breaks other use cases while solving this issue.

larshp:
hmm, yea, looks like there are some issues with the fix

Martin1994:
Hi @nicoespeon. Sorry for the delay. I've fixed the arrow bug. Please review the latest commit.

However, for the unexpected path on the ```develop``` branch, I think it would be a "won't fix". In the test code, the ```develop``` branch is branched before the very first commit, which should be an unexpected behaviour.

nicoespeon:
@Martin1994 thanks for fixing issues.

I checked and confirm that the issue with the develop branch is legit regarding the code. It worked by chance but actually shouldn't since there is no initial commit. We might deal with it forcing an initial commit on graph creation, but that's out of this PR scope.

I just tested your code and it works like a charm without noticed regression.

👍

PR #91, opened 2015-12-27
Add examples from master to GH-Pages?

Potherca:
I thought it _might_ be useful to have the examples in the `gh-pages` branch so they can be seen in action.

As creating a PR took about as much effort as creating a ticket, this seemed more effective.

This PR can be seen in action at http://potherca-contrib.github.io/gitgraph.js/

This might cause more work for the maintainers (the examples might need to be updated in the future), so I can understand if this may _not_ be deemed useful.

Thoughts?


nicoespeon:
Well that's a great idea. Examples are here already, let's use them on gh-pages.

I don't know while I didn't get back to you earlier. But there you go.
Thanks 👍

PR #93, opened 2016-01-16
[BUGFIX]

duahcram:
Cannot see tooltip on commit with FireFox 43.
Error message was "TypeError: setting a property that has only a getter" on event.offsetX.
Just remove part of FFi MouseEvent fix line 336 and 337.


nicoespeon:
Hi @duahcram, thanks for pointing out.

I'll check all that this weekend, I'd have some free time =)

nicoespeon:
Hi @duahcram, thanks for the fix, it works like a charm. Thanks for catching this :+1: 

Thanks @Davilink for your proposal too. I'll release a patch version with this fix.

Issue #103, opened 2016-03-25
Tag alignment 


Jayde-Im:
Your work is great, and I appreciate using it. 

But one little part is bothering me. 
When the font sizes of the tag and the message are different and tag name is long, the tag is not properly aligned.

I'm using this template in the latest version of gitgraph.js. (Also, latest version of Chrome)
```
var graphConfig = new GitGraph.Template({
  'colors': [ "#979797", "#008fb5", "#f1c109" ],
  'commit': {
    'message': {
      'displayBranch': false
    }
  }
});
```

And this is an example,
I apply this template to gitflowsupport.js, it looks like below.

![image](https://cloud.githubusercontent.com/assets/13883550/14044636/2544630a-f2d6-11e5-95c7-3e8365481e93.png)

The "v1.0.2-a-super-tag" tag isn't aligned with others.

Yeah, I know it's quite minor issue, but it's not pretty in my project. 
Although, this library is really good! :smile:

nicoespeon:
Hi @Jayde-Im!

Thanks a lot for spotting out this and for the PR! 

I'll have a look, test it and get it fixed quickly ;-)

PR #104, opened 2016-03-25
[Bug Fix] Align tags properly

Jayde-Im:
#103

Cause: 

- Tag's textWidth is measured in message's font-size instead of tag's font-size.

Changes:

- Apply tag's font-size to context before measuring tag's textWidth.
- Apply message's font-size to context after drawing tag.

You can test it on gitflowsupport.js.

### DIFF
--- a/src/gitgraph.js
+++ b/src/gitgraph.js
@@ -939,12 +939,11 @@
       this.arrow();
     }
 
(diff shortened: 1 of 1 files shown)


nicoespeon:
It's all good, thanks again @Jayde-Im :+1:

PR #106, opened 2016-05-06
Restore line dash after applying it to branch

jru:
Changing the line dash for a branch propagates it to other objects
like commits or tags that do not explicitly set it before start
drawing.

Prevent it by saving the dash setting before being overridden and
restore it afterwards.


nicoespeon:
That’s a good catch, thanks!

Issue #109, opened 2016-07-15
Add the possibility to change the initial CommitOffsetX and CommitOffsetY


bpatra:
We may encounter situations where the gitgraph spreads outside the canvas.  Especially when we change the size of the commit dots.
See this example
![outside](https://cloud.githubusercontent.com/assets/2801702/16882052/58f890d8-4abd-11e6-8a6a-1e04319b2e6c.jpg)

This could be fix if we override with settings options the initial values of 
`commitOffsetX` and `commitOffsetY`

PR #110, opened 2016-07-15
Support offset in options

bpatra:
This PR attempts to fix issue #109


bpatra:
Bad PR because it tries to merge with develop while develop is behind master

Issue #113, opened 2016-07-25
Add the possiblity to avoid the black borders


bpatra:
For tagging branch, the font text has necessarily a black font color and is put in a rectangle with black border.

![withborder](https://cloud.githubusercontent.com/assets/2801702/17109721/825fe398-5299-11e6-8111-3c17f046a18d.jpg)

It would nice if we can ignore the rectangle and just display the text with the commit color.

![noborder](https://cloud.githubusercontent.com/assets/2801702/17109759/a6f497bc-5299-11e6-8bb0-867d3553eacf.jpg)

PR #114, opened 2016-07-25
Add the possibility to avoid the black border for tag rendering. (solution for issue  #113)

bpatra:
If  noTagBlackBorder option is set to true we use the commit color for the tag text otherwise (previous behavior) we use a black font and the commit color for background.

This should fix the issue (enhancement)  #113

### DIFF
--- a/src/gitgraph.js
+++ b/src/gitgraph.js
@@ -968,6 +968,7 @@
     this.tag = options.tag || null;
     this.tagColor = options.tagColor || options.color;
     this.tagFont = options.tagFont || this.template.commit.tag.font;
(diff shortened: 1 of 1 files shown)


nicoespeon:
Great @bpatra!

That's awesome. I'll take a look at it as soon as possible − coming back from holidays, this is a nice surprise.

Thanks.

nicoespeon:
Just reviewed and tested your PR, it works well.

I'm accepting it right now. I will add doc & illustrate this into examples too.
Also, I will certainly rename the options to opt for a positive assertion defaulted to `true` because `noTagBlackBorder: false` could be a bit confusing I think. Not a big deal though.

Again: thanks a lot =)

PR #116, opened 2016-08-23
Fix offsetX and offsetY for older FF

Fisjkars:
Hello,

I had an issue when the canvas is not on the top of the page. (Firefox 38.XX)
I change these two lines in order to be compatible with old Firefox because according to the definition of layerY.

_This property takes scrolling of the page into account, and returns a value relative to the whole of the document, unless the event occurs inside a positioned element, **where the returned value is relative to the top left of the positioned element.**_

pageY take in account the whole page.

Thanks for your works!

### DIFF
--- a/src/gitgraph.js
+++ b/src/gitgraph.js
@@ -374,8 +374,18 @@
    **/
   GitGraph.prototype.applyCommits = function ( event, callbackFn ) {
     // Fallback onto layerX/layerY for older versions of Firefox.
-    var offsetX = event.offsetX || (event.pageX - $("#" + this.elementId).offset().left);
-    var offsetY = event.offsetY || (event.pageY - $("#" + this.elementId).offset().top);
+    function getOffsetById(id) {
+            var el = document.getElementById(id);
+            var rect = el.getBoundingClientRect();
+
+            return {
+                top: rect.top + document.body.scrollTop,
+                left: rect.left + document.body.scrollLeft
+            };
+        }
+
+    var offsetX = event.offsetX || (event.pageX - getOffsetById(this.elementId).left);
+    var offsetY = event.offsetY || (event.pageY - getOffsetById(this.elementId).top);
 
     for ( var i = 0, commit; !!(commit = this.commits[ i ]); i++ ) {
       var distanceX = (commit.x + (this.offsetX + this.marginX) / this.scalingFactor - offsetX);
--- a/src/gitgraph.js
+++ b/src/gitgraph.js
@@ -374,8 +374,8 @@
    **/
   GitGraph.prototype.applyCommits = function ( event, callbackFn ) {
     // Fallback onto layerX/layerY for older versions of Firefox.
-    var offsetX = event.offsetX || event.layerX;
-    var offsetY = event.offsetY || event.layerY;
+    var offsetX = event.offsetX || (event.pageX - $("#" + this.elementId).offset().left);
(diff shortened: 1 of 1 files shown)


nicoespeon:
Hi @Fisjkars and thanks for that =)

That seems cool, excepted that Gitgraph has no dependency on jQuery. So this wouldn't work unless you have jQuery loaded before Gitgraph by chance.

Could you implement this one without the use of jQuery then?

Thanks a lot =)

Fisjkars:
Hello,

oh yes... i didn't saw that point :/
I had a look on jquery api code for offset to have an equivalent cross-client solution for these functions, unfortunately, imho I think it's not a sustainable solution for gitgraph to integrate this code. 

Btw, i'm not a js/jquery expert, so if you have alternative solution, i will happy to implement it.

You will find enclosed the concerned code in jQuery API.

thanks again for your work, have a nice day.

`jQuery.offset = {
	setOffset: function( elem, options, i ) {
		var curPosition, curLeft, curCSSTop, curTop, curOffset, curCSSLeft, calculatePosition,
			position = jQuery.css( elem, "position" ),
			curElem = jQuery( elem ),
			props = {};

		// Set position first, in-case top/left are set even on static elem
		if ( position === "static" ) {
			elem.style.position = "relative";
		}

		curOffset = curElem.offset();
		curCSSTop = jQuery.css( elem, "top" );
		curCSSLeft = jQuery.css( elem, "left" );
		calculatePosition = ( position === "absolute" || position === "fixed" ) &&
			( curCSSTop + curCSSLeft ).indexOf( "auto" ) > -1;

		// Need to be able to calculate position if either
		// top or left is auto and position is either absolute or fixed
		if ( calculatePosition ) {
			curPosition = curElem.position();
			curTop = curPosition.top;
			curLeft = curPosition.left;

		} else {
			curTop = parseFloat( curCSSTop ) || 0;
			curLeft = parseFloat( curCSSLeft ) || 0;
		}

		if ( jQuery.isFunction( options ) ) {

			// Use jQuery.extend here to allow modification of coordinates argument (gh-1848)
			options = options.call( elem, i, jQuery.extend( {}, curOffset ) );
		}

		if ( options.top != null ) {
			props.top = ( options.top - curOffset.top ) + curTop;
		}
		if ( options.left != null ) {
			props.left = ( options.left - curOffset.left ) + curLeft;
		}

		if ( "using" in options ) {
			options.using.call( elem, props );

		} else {
			curElem.css( props );
		}
	}
};`

nicoespeon:
Hi @Fisjkars =)

Could you tell me if that works for you if you replace:

```javascript
var offsetX = event.offsetX || (event.pageX - $("#" + this.elementId).offset().left);
var offsetY = event.offsetY || (event.pageY - $("#" + this.elementId).offset().top);
```

 with following code:

```javascript
function getOffsetById(id) {
  var el = document.getElementById(id);
  var rect = el.getBoundingClientRect();
  
  return {
    top: rect.top + document.body.scrollTop,
    left: rect.left + document.body.scrollLeft
  };
}

var offsetX = event.offsetX || (event.pageX - getOffsetById(this.elementId).left);
var offsetY = event.offsetY || (event.pageY - getOffsetById(this.elementId).top);
```

Fisjkars:
Hello @nicoespeon 
This code is working fine on my configuration (IE11, FF38.8.0, FF24.8.1 & FF10.0.12).
Thanks for the support.

nicoespeon:
Thanks for the contribution, I'll release this fix 👍

PR #122, opened 2016-10-01
Update link to Jasmine

mortonfox:


nicoespeon:
Thanks for catching that @mortonfox 👍

PR #125, opened 2016-11-19
Improve readability of markdown files as plaintext

Graham42:
When reading in a terminal, it's much nicer to have shorter line lengths. 

In this change I wrapped lines at 100 characters for the most part, and moved link markup to the end of paragraphs.

Unfortunately GitHub doesn't give a very nice diff of what changed, but using the `--word-diff` option in the terminal should ignore the wrapping changes


nicoespeon:
Hi @Graham42 and thanks for this improvement 👍

PR #130, opened 2016-12-08
Issue: Click object is not passed to the consumer end click handler

Giridharan2112:
Fix: Pass the event object to the handler along with the commit object


nicoespeon:
Hi @Giridharan2112!

Just tested your implementation and it works like a charm. Thanks 👍

PR #133, opened 2016-12-26
Added fix for JSHint via Grunt (null path warning).

cgddrd:
Experienced the following warning whilst attempting to run JSHint via Grunt:

`Warning: Path must be a string. Received null Use --force to continue.`

causing Grunt to fail unless forced to continue. 

This PR applies a workaround suggested as part of the original issue raised for JSHint: 

[https://github.com/jshint/jshint/issues/2922](https://github.com/jshint/jshint/issues/2922)

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -116,7 +116,8 @@ module.exports = function ( grunt ) {
       src: [ "src/*.js" ],
       options: {
         force: true,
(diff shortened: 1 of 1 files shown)


nicoespeon:
Well, the fix looks good to me. If it solves an issue you had, it will be helpful to others too.

Thanks 👍

PR #135, opened 2016-12-30
fix typo in README.md

yodalee:
It seems that git@ becomes git://


nicoespeon:
Thanks for catching this 👍

PR #146, opened 2017-03-07
Support updating branches from parent

ewirch:
let master = gitgraph.branch("master");
master.commit(" ");
let branch = master.branch("branch");
branch.commit(" ");
master.commit(" ");
// update "branch", failed previously:
master.merge(branch);

### DIFF
--- a/src/gitgraph.js
+++ b/src/gitgraph.js
@@ -856,9 +856,15 @@
     commitOptions.type = "mergeCommit";
     commitOptions.parentCommit = _getParentCommitFromBranch(this);
 
-    var branchParentCommit = firstBranchCommit.parentCommit;
-    var targetBranchParentCommit = _getParentCommitFromBranch(targetBranch);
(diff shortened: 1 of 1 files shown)


nicoespeon:
Hi @ewirch and thanks for opening a PR 😃 

I'll give it a look by the end of the week - probably this week-end actually.

nicoespeon:
Hi @ewirch!

I'm giving a look at this and I have trouble figuring out the original issue.

Using the code snippet you gave in the description, here is the result before your PR:

before

And here is the result after:

after

Am I missing something? Maybe you could make a fiddle to illustrate the issue?

ewirch:
Hm. Not reproducable any more. I must have been using a older version and applied the patch blindly to the current version. This line:

```
var isFastForwardPossible = (branchParentCommit && branchParentCommit.sha1 === targetBranchParentCommit.sha1);
```
did not have the null check in the version I've been using. Just forget the PR.

nicoespeon:
Thanks for opening the PR. Even if it was already fixed in the latest version, I appreciate the intent 👍 

Have a nice week 😉

Issue #140, opened 2017-01-11
Graph with arrows doesn't render properly multiple branches from the same commit


nicoespeon:
<img width="276" alt="arrow-commit-pushed-from-same-branch" src="https://cloud.githubusercontent.com/assets/1094774/21838794/7a1ea898-d7d4-11e6-85db-174fef4d52e3.png">

Arrow is expecting the branch dash to come vertically since commit is "pushed in Y".

Source code:

```javascript
var gitgraph = new GitGraph({
  template: "blackarrow",
});

var master = gitgraph.branch("master")
master.commit();

var branch0 = master.branch("branch0")
branch0.commit();
master.commit();
branch0.commit();

var branch1 = branch0.branch("branch1")
master.merge(branch0)

var branch2 = branch1.branch("branch2")
var branch3 = branch1.branch("branch3")
branch3.commit().commit();
branch2.commit();
branch3.commit();
branch2.commit();
```

PR #150, opened 2017-03-28
fix issue 140 wrong arrow angle

yodalee:
#140 

The steps to reproduce error:

var master = gitgraph.branch("master");
master.commit();
var branch0 = master.branch("branch0");
var branch1 = branch0.branch("branch1");
var branch2 = branch0.branch("branch2");
var branch3 = branch0.branch("branch3");

branch1.commit()
branch2.commit()
branch3.commit()

The arrow point from branch0 to branch2, branch3
will have vertical directions.

The cause:
There is a property 'isForkCommit' when drawing arrow
which if set to true, the direction of arrow will
re-calculate the vertical if difference in Y direction
is larger than Y spacing of template.
The direction of branch1 arrow is correct since
difference in Y direction is equal Y spacing of template.

The fix:
The property 'isForkCommit' seems unused.
Simply remove the property from calculation of arrow direction
can solve the issue.
However, more test on other cases are required to ensure that
removing isForkCommit property not affect other render.

### DIFF
--- a/examples/index.js
+++ b/examples/index.js
@@ -32,7 +32,7 @@ var myTemplate = new GitGraph.Template(myTemplateConfig);
  ***********************/
 
 var config = {
-  template: "metro", // could be: "blackarrow" or "metro" or `myTemplate` (custom Template object)
+  template: "blackarrow", // could be: "blackarrow" or "metro" or `myTemplate` (custom Template object)
   reverseArrow: false, // to make arrows point to ancestors, if displayed
   orientation: "vertical",
   // mode: "compact" // special compact mode: hide messages & compact graph
@@ -53,7 +53,7 @@ gitGraph.commit("Initial commit");
 gitGraph.commit("My second commit").commit("Add awesome feature");
 
 // Create a new "dev" branch from "master" with some custom configuration
-var dev = gitGraph.branch({
+var dev = master.branch({
   name: "dev",
   color: "#F00",
   commitDefaultOptions: {
@@ -77,6 +77,10 @@ var commitConfig = {
   author: "Me <me@planee.fr>"
 };
 gitGraph.commit(commitConfig);
+
+// Create another from "master"
+var feature3 = master.branch("feature3")
+feature3.commit().commit();
 
 /***********************
  *      CHECKOUT       *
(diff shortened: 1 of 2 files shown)


nicoespeon:
Thanks for helping @yodalee, that's really nice of you 👍 

It sounds good, but I realized the project example doesn't illustrate these use cases.
I added 2 use cases so we can see the problem spotted in #140. 

I pushed the commit on your branch so you can base your work on it: https://github.com/yodalee/gitgraph.js/commit/d3723cce5dc6fc2320f7f68b7d1117773cbf429d
(just run `grunt server` to see the graph in your browser)

Unfortunately, you can also see that this fix is causing the same trouble for another use case − which was using the `isForkCommit` variable:

capture d ecran 2017-03-29 a 23 38 38

Our issue here is that we don't handle the case of multiple branches created from a branch that has no commit. This case should behave just like if the origin branch had commits: the line should draw an angle, going straight to the commit − see picture above.

This happens because "the angle" only appears when commits of the other branches are "pushed" in Y because of the commit on the origin branch.
But this (= the angle) should also happen when the origin branch has no commit.

yodalee:
Yes now there is use case, I can do more research on this problem.

yodalee:
I think the problem lies in the branch line direction. In your case, the line become segment and a straight line in the empty branch case. Need to find why.

yodalee:
I think I found a better fix to the problem, here is the commit

nicoespeon:
Hi @yodalee!

I tested your proposal and it works nicely.
I found rendering correct, especially regarding what it looks like before your fix. I tried to change the template spacing and I've got issues with other parts of rendering way before having troubles with the segment + arrows here 😉 

Thanks a lot for contributing 👍

nicoespeon:
I spoke a little too quickly. After checking issue #140 I realized that there is a regression.

If you take this snippet:

```javascript
var gitgraph = new GitGraph({
  template: "blackarrow",
});

var master = gitgraph.branch("master")
master.commit();

var branch0 = master.branch("branch0")
branch0.commit();
master.commit();
branch0.commit();

var branch1 = branch0.branch("branch1")
master.merge(branch0)

var branch2 = branch1.branch("branch2")
var branch3 = branch1.branch("branch3")
branch3.commit().commit();
branch2.commit();
branch3.commit();
branch2.commit();
```

The rendering is broken:

capture d ecran 2017-04-03 a 09 00 13

I re-open the issue and will wait to fix that point before releasing it.

yodalee:
Yes we need some test case to verify that this modification is correct and will not cause other problem.

nicoespeon:
I updated the example to illustrate the use case: https://github.com/nicoespeon/gitgraph.js/commit/8f61ffc23dbe2b0b52759696e1252de750c180c0

I also reverted your change temporarily so I can release new versions of the graph (3d62564d9698945818ef3d4f0cbe38abf686c08e).

I keep #140 open until we figure out how to fix the issue without regression.

PR #154, opened 2017-04-05
Issue140 vertical arrow with multiple branch commit

yodalee:
The steps to reproduce error:

var master = gitgraph.branch("master");
master.commit();
var branch0 = master.branch("branch0");
var branch1 = branch0.branch("branch1");
var branch2 = branch0.branch("branch2");

branch1.commit()
branch2.commit()

The arrow point from branch0 to branch2, branch3
will have vertical directions.

The cause:
When create a new commit object, and the commit is the first commit
to this branch, it will first check parentBranch has commit or not.
If it has commit, it will push a new point to path to create a joint
to the path, in order to create more beautiful path.
In the above case, the branch0 has no commit when branch1, 2
commit, so branch1, 2 will have no breakpoint in path, instead
the line will straightly draw from master to the commit.

The fix:
The check of parentCommit is empty or not is actually not needed.
We can always draw the joint.
The position of the joint is calculated as follow:
x equals to commit x, y equals to startpoint y minus template spacing

So if difference of y is larger than template spacing, it will create
a joint at one template spacing then straight line to new commit.

TODO:
Currently the angle of arrow is calculated by various condition.
It will check it is forkcommit or not, then determines that if
it is vertical or not.
I think a simpler way is to calculate angle of arrow by path.
First find the point in path that x, y equal to commit.
Then calculate angle by x,y difference of previous point and current
point

### DIFF
--- a/src/gitgraph.js
+++ b/src/gitgraph.js
@@ -768,19 +768,13 @@
 
     if (!isFirstBranch && isPathBeginning) {
       this.pushPath(this.startPoint);
-
-      // Trace path from parent branch if it has commits already
-      if (this.parentBranch.commits.length > 0) {
-        this.pushPath({
-          x: this.startPoint.x - this.parentBranch.offsetX + this.offsetX - this.template.commit.spacingX,
-          y: this.startPoint.y - this.parentBranch.offsetY + this.offsetY - this.template.commit.spacingY,
-          type: "join"
-        });
-
-        var parent = _clone(this.startPoint);
-        parent.type = "join";
-        this.parentBranch.pushPath(parent);
-      }
+      // Add a path point to startpoint + offset - template spacing
+      // So that line will not go through commit of other branches
+      this.pushPath({
+        x: commit.x,
+        y: this.startPoint.y - this.template.commit.spacingY,
+        type: "joint"
+      });
     } else if (isPathBeginning) {
       point.type = "start";
     }
--- a/src/gitgraph.js
+++ b/src/gitgraph.js
@@ -770,11 +770,19 @@
       this.pushPath(this.startPoint);
       // Add a path joint to startpoint + template spacing
       // So that line will not go through commit of other branches
-      this.pushPath({
-        x: commit.x,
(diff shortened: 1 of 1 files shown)


yodalee:
Except the testcase in examples directory, I test following testcase:
with empty branch:
```
var master = gitgraph.branch("master");
master.commit();
var branch0 = master.branch("branch0");
var branch1 = branch0.branch("branch1");
var branch2 = branch0.branch("branch2");

branch2.commit()
branch1.commit()
```

without empty branch
```
var master = gitgraph.branch("master");
master.commit();
var branch0 = master.branch("branch0");
var branch1 = master.branch("branch1");

branch0.commit()
branch1.commit()
```

Test with merge:
```
var master = gitgraph.branch("master")
master.commit();

var branch0 = master.branch("branch0")
branch0.commit();

var branch1 = branch0.branch("branch1")
master.merge(branch0)

var branch2 = branch1.branch("branch2")
var branch3 = branch1.branch("branch3")
branch3.commit();
branch2.commit();
branch1.commit()
```

So far all act as expected.

nicoespeon:
Hey, thanks for this @yodalee!

I tested it on the different examples we have, and it perfectly solve the issue.
But a regression appears if you choose another orientation for the graph than "vertical". For example, choosing "horizontal":

```js
var gitGraph = new GitGraph({
  template: "blackarrow",
  orientation: "horizontal"
});
```

Before these changes:

capture d ecran 2017-04-12 a 08 02 51


After these changes:

capture d ecran 2017-04-12 a 08 02 38

I think there might be a calculation to take into account when we push the "joint" point to consider we might not be vertically aligned.

yodalee:
Sorry for the late response, I just got busy last month.
I just push a fix to this pull request that distinguish vertical and horizontal orientation. Hope this will work.

nicoespeon:
Thanks @yodalee, it seems to work perfectly with all orientations.

I'll just refactor the code to use the existing helper `_isVertical(graph)`, no big deal.
Thanks again for your contribution. I'm merging this and I'll make a new release to close #140 👍 🎉

PR #158, opened 2017-05-04
Change tag frame to look like a tag

RobotCaleb:
I did this for a workflow I was modeling. I liked it, so I'm throwing at you lot to see if it works for your goals. :)

I tried to make the change not have any external impact to current users. A good extension to this feature would probably be to expose a `frameType` to the commit and tag options.

Old:
oldtags

New:
newtags

### DIFF
--- a/src/gitgraph.js
+++ b/src/gitgraph.js
@@ -1138,7 +1138,7 @@
           this.labelColor,
           this.labelFont,
           this.template.branch.labelRotation,
-          true);
+          "box");
       } else {
         _drawTextBG(this.context,
           this.x + this.template.commit.spacingX,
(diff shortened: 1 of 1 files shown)


nicoespeon:
Hi @RobotCaleb and thanks a lot for this PR. 

That's a nice cosmetic feature 👍 

I'd just have one request though: keep the default tag box at none and make it opt-in with something like `frameType`, as you said. Or make it the default, but give the ability to users to fall-back to the current tag box.

This is preferable as the tag looks good for `vertical` orientation only, but not the others (`vertical-reverse`, `horizontal` and `horizontal-reverse`).

Thanks 😃

Issue #160, opened 2017-05-06
Branch labels are shown on first commit in branch, not head of branch


aspiers:
I see that this is [a known issue](https://github.com/nicoespeon/gitgraph.js/issues/142#issuecomment-277069476), however I think it's important enough that it merits being tracked in its own issue.  I also appreciate (as that comment points out) that there are some fundamental design flaws which would need to be addressed to fix this properly.  Nevertheless it appears possible to at least work around this issue (PR to follow shortly).

Issue #142, opened 2017-01-19
Mode: Branch pointers only


JKrag:
It would be lovely if there was a mode "between" compact and standard. I would like something that just adds branch labels to the commit currently representing a branch.

Something along the lines of the following picture: 
![image](https://cloud.githubusercontent.com/assets/1712950/22108196/2b6303dc-de52-11e6-9b5e-72a44f46096b.png)
(generously borrowed from https://www.atlassian.com/git/tutorials/using-branches).

JKrag:
I played around and noticed that there is a showLabel feature. As I am not used to JS, It took me some time even to work out how to set this property without having to write a whole template myself.
When I finally got it working, I found out that it didn't do what I hoped anyway :-)

I have now spent the evening reading and hacking around in the code and learnt a lot of JS along the way too. I did manage to put together a crude proof of concept: http://jsfiddle.net/fqpfgt9c/

If anyone wants to see exactly what I changed, the diff can be seen here: https://github.com/JKrag/gitgraph.js/commit/9b85562d731384c40d3f7a3938ef5dfa8354db42 

There is still a long way to go to figure out the spacing and stuff, but my hope is that someone more experienced in the codebase and JS can step in and build on this idea. 

I am really hoping to also get support for multiple branch labels at the same commit, and an optional HEAD label that shows the current checkout, but those are topics for other/future feature requests.

I.e. The end goal is that I want to be able to produce gitgraph.js graphs that are functionally equivalent to the following minimal example:
<img width="198" alt="cursor_and_explain_git_with_d3" src="https://cloud.githubusercontent.com/assets/1712950/22131531/5bb08002-deb4-11e6-834a-cc577378c8fa.png">
(Generated at [Explain git with d3](http://onlywei.github.io/explain-git-with-d3/#zen) using the commands:
```
git branch foo
git branch bar
git commit
git checkout bar
```

As stated, I do consider these separate features:
 -  This issue: branch labels/pointers at branch head
 - Multiple labels at same commit
 - HEAD label at current checked out branch 

I plan to create separate issues for the last two, but thought the overall idea was worth mentioning here as it may influence on implementation choices.

I hope this all makes a bit of sense and doesn't conflict with the current path of the project.

nicoespeon:
Hi @JKrag,

Thanks for this detailed issue with many suggestions and digging.

Concerning the "display branch tags in compact mode" part, there is no obvious / documented way to do that indeed.
[...]

PR #161, opened 2017-05-06
show branch label on last commit in branch, not first

aspiers:
Forward-ported from JKrag@9b85562 which was found here:

https://github.com/nicoespeon/gitgraph.js/issues/142#issuecomment-273948476

Provides a partial fix or workaround for #160.

### DIFF
--- a/src/gitgraph.js
+++ b/src/gitgraph.js
@@ -744,11 +744,12 @@
     var isFirstBranch = !(options.parentCommit instanceof Commit);
     var isPathBeginning = this.path.length === 0;
 
-    options.showLabel = (isPathBeginning && this.showLabel);
-    if (options.showLabel) {
-      options.x -= this.template.commit.spacingX;
-      options.y -= this.template.commit.spacingY;
-    }
+    var cmts = options.branch.commits;
+    cmts.forEach(function (commit) {
+      commit.showLabel = false;
+    });
+
+    options.showLabel = (this.showLabel);
 
     var commit = new Commit(options);
     this.commits.push(commit);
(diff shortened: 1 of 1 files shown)


nicoespeon:
Hi @aspiers and thanks for pushing this 👌 

There is a lot of commented out code however. Could you clean it up before I test this PR?

Thanks a lot for helping 😃

aspiers:
@nicoespeon Done.

nicoespeon:
Thanks @aspiers.

I found a regression with following snippet:

```js
var metroWithBranchLabels = new GitGraph.Template({
  colors: ["#979797", "#008fb5", "#f1c109"],
  branch: {
    lineWidth: 10,
    spacingX: 50,
    labelRotation: 0,
    showLabel: true
  },
  commit: {
    spacingY: -80,
    dot: {
      size: 14
    },
    message: {
      font: "normal 14pt Arial"
    }
  }
});

var gitGraph = new GitGraph({
  template: metroWithBranchLabels,
  orientation: "horizontal",
  mode: "compact"
});

var master = gitGraph.branch("master");
master.commit();

var dev = master.branch("dev");
dev.commit()
   .merge(master)
   .commit();

master.commit();
```

Before:

capture d ecran 2017-05-12 a 09 32 12

After:

capture d ecran 2017-05-12 a 09 29 39

Could you give it a look?

aspiers:
In terms of label positioning, "After" is more correct than "Before".  I guess you are referring to the extra parent of `dev^`?  That does indeed look like a regression.  However my understanding based on [this comment](https://github.com/nicoespeon/gitgraph.js/issues/142#issuecomment-277069476) is that the core of the library is fundamentally the wrong design and needs a rewrite in order to model git's data structures correctly.  Without doing that I'm not sure it's worth spending much effort on trying to fix smaller issues, and sadly I don't have the time to do that anyway :-(

nicoespeon:
In terms of label positioning, you're right, and this was the point of the PR indeed 👌 

And yup, I was referring to the extra parent that appears and looks weird.

Thanks a lot for trying to help with this one. But as you noticed in my other comment, the design of the library needs to be rewrote. I don't think it worth spending more time to find complex workarounds until then for this kind of issues. It'd better pay off to make things cleaner from the root.

I'm closing this PR then. Still, I really appreciate your contribution 😃 

I'll tackle this when I'll have enough time / motivation with @fabien0102 to make a v2 of the lib, with a proper design 😉

PR #163, opened 2017-05-23
Adding incompatible version to showcase using support branches

cardil:
Adding incompatible version to showcase using support branches with them.

Also I change the merge of release to be proper - develop should have a tagged master merged to it, so `git describe` work properly.

PS. I like G.M. :-P


nicoespeon:
Hi @cardil,

Thanks for improving the gitflow showcase and spotting the mistake with the merge back to `develop` after release is done 👍

PR #165, opened 2017-05-25
Ff 53 fix

### LINKED ISSUE #164: FF53 no branch label tooltips
Tooltips are currently broken and not showing up in FireFox 53. The console gives an error on hover of:
`TypeError: setting a property that only has a getter`
This is due to trying to set event.x for FireFox 52 and before which used event.[xy] for the screen coordinates and event.[XY] for client(canvas) coordinates.

moore3071:
Fixes #164 and should work for all browsers by using event.clientX and event.clientY. (https://developer.mozilla.org/en-US/docs/Web/Events/mouseover)

### DIFF
--- a/src/gitgraph.js
+++ b/src/gitgraph.js
@@ -375,14 +375,8 @@
         return;
       }
 
-      // Fix firefox MouseEvent
-      if (typeof InstallTrigger !== "undefined") /* == (is Firefox) */ {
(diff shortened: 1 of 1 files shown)


nicoespeon:
Hey, thanks @moore3071!

I tested it and it works like a charm 👍

Issue #159, opened 2017-05-05
Graph rendering fails with basic branching


laymain:
Hi,

There's a bug on the graph rendering when a commit is made on a branch
after it has been used to create another branch:

![image](https://cloud.githubusercontent.com/assets/389065/25738233/3052e39e-317c-11e7-8fdb-97adff269c63.png)

```javascript
var graph1 = new GitGraph({
  elementId: 'badGraph',
  mode: 'compact',
  template: 'blackarrow'
});
var m1 = graph1
  .branch("master")
  .commit();
var r1 = m1
  .branch('1.0.x');
m1.commit();
r1
  .commit()
  .merge(m1);
```

If we do the commit on the new branch before the commit on the first one, it renders finely:

![image](https://cloud.githubusercontent.com/assets/389065/25738270/64b4a7f8-317c-11e7-9f6b-ab2c2b43abab.png)

```javascript
var graph0 = new GitGraph({
  elementId: 'workingGraph',
  mode: 'compact',
  template: 'blackarrow'
});
var m0 = graph0
  .branch("master")
  .commit();
var r0 = m0
  .branch('1.0.x')
  .commit();
m0.commit();
r0.merge(m0);
```

https://jsfiddle.net/Laymain/k0z0277g/

nicoespeon:
Hi @laymain, thanks for pointing out this issue, this is definitely a bug!

yodalee:
I think this is related to the PR I push, I will dig into this issue.

yodalee:
I take some time to trace the bug, at least in tag v1.1.0, the output graph is still OK, I will use bisect to see where goes wrong.

yodalee:
Rough tracing takes me to cc8110f04971f95ac90833ff65bab00181fe84bb
This commit cause the arrow start at first commit

nicoespeon:
Hi @laymain,

Thanks to @yodalee, this should be fixed with v1.11.4 (https://github.com/nicoespeon/gitgraph.js/releases/tag/v1.11.4).

laymain:
Hi,

The wrong arrow angle is fixed, that's nice.
But is there any reason why it does not render in the same way
if a commit is made before or after branching?

![image](https://user-images.githubusercontent.com/389065/29109585-d0588d5a-7ce3-11e7-946b-ae2e841f1acc.png)

nicoespeon:
[...]

PR #171, opened 2017-07-22
fix wrong arrow angle on branching

yodalee:
Relate Issue:
Issue #159

The Problem:
The following source code will create wrong branch arrow.
```
var graph1 = new GitGraph({
  elementId: 'badGraph',
  mode: 'compact',
  template: 'blackarrow'
});
var m1 = graph1.branch("master");
m1.commit();
var r1 = m1.branch('1.0.x');
m1.commit();
r1.commit();
```
The arrow pointed to r1 commit will pointed from second commit of m1
not follow the line from first commit of m1.

The Cause:
When create a commit, if user does not specify parent commit in options
gitgraph.js will execute following code:
```
options.parentCommit = options.parentCommit || _getParentCommitFromBranch(this);
```
to determine the parent commit of the created commit.
In the function `_getParentCommitFromBranch`, it will try to get the last
commit from current branch, then from parent branch. Since `r1.commit()`
is the first commit in its branch `1.0.x`, it will set its parent commit
to last commit of its parent branch, that is `master`. And since master
already add some new commit, the arrow will point from wrong commit, the
last commit of `master` branch.

There is code that deal with branch case:
```
if (options.parentCommit instanceof Commit === false && this.parentBranch instanceof Branch) {
  options.parentCommit = this.parentCommit;
}
```
It set `parentCommit` to the correct commit. However, since
`options.parentCommit` already been set by previous code, this code will
never execute since `options.parentCommit` is always a Commit.

The fix:
The fix is combining two scenario. If current branch already has commits,
the parentCommit will be set to last commit in this branch. Otherwise,
it will set to parent commit of this branch, which will be the correct commit
instead of last commit of parent branch.

### DIFF
--- a/src/gitgraph.js
+++ b/src/gitgraph.js
@@ -673,7 +673,17 @@
       this.template.commit.color ||
       this.template.colors[columnIndex];
     options.parent = this.parent;
-    options.parentCommit = options.parentCommit || _getParentCommitFromBranch(this);
+
+    // Set parentCommit
+    // If there is commit in this branch, set parentCommit to last commit of this branch
+    // otherwise, set parentCommit to this.parentCommit, the start point of this branch
+    if (!options.parentCommit) {
+      if (_getLast(this.commits)) {
+        options.parentCommit = _getLast(this.commits);
+      } else {
+        options.parentCommit = this.parentCommit;
+      }
+    }
 
     // Special compact mode
     if (this.parent.mode === "compact" &&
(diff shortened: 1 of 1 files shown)


nicoespeon:
Hi @yodalee 👋 

Coming back from holidays, that's a wonderful PR and a very nice debug. Thanks a lot 👍 🎊 

I just tested your change, it works well and I agree with the debug you nicely described − thanks for that, it helps a lot.

Issue #187, opened 2018-01-21
Gitgraph.js v2 - The big remodelling!… and more


fabien0102:
# Gitgraph.js v2

**Project board to track completion:** https://github.com/nicoespeon/gitgraph.js/projects/2

No timeline for a release yet. But…
@fabien0102 & @nicoespeon work at least 1h of remote pair-programming every week on this 👨‍🎤🎸

### When everything else will be done

- [x] List changes between v1 and v2 (may find missing stuff for v2)
- [x] Clean public API
- [x] Update README
- [x] Publish an alpha-release of `gitgraph-react`

### To fix for release

- [x] Adjust canvas SVG dynamically
- [x] Fix deleted branch scenario (for imports)
- [x] Fix merged branch which doesn't appear scenario
- [x] Implement missing fast-forward

### General

- [x] Refactor the repository with `/packages/*` pattern
- [x] Build stack (rollup, uglify, typescript…)
- [x] Fix travis build (setup yarn)
- [x] Fix Prettier config on different packages & run as pre-commit hook
- [x] Jest at the root level (lerna)
- [x] tslint at the root level (lerna) with `tslint --fix` as pre-commit hook

### Gitgraph Core

- [x] Bootstrap (with almost the same API as the legacy gitgraph)
- [x] Implement branches colors
- [x] Orientations
- [x] Compact mode
- [x] Tags
- [x] `templateExtend()` to easily create new templates from existing ones
- [x] Template metro
  - [x] Bézier curves
  - [x] Branch colors disappear after 3 branches (doesn't loop back to first color)
- [x] Template blackarrow
  - [x] Circles (inner colors)
  - [x] Arrows
  - [x] Reverse arrows
- [x] Import from a defined format (e.g. git2json output)
  - [x] Multiple commits, one branch
  - [x] Multiple branches
  - [x] Color / style (add tests)
  - [x] Tags
  - [x] JOI to validate input data

### Gitgraph React (rendering)

- [x] Bootstrap (first gitgraph-core integration test)
  - [x] Set up the workspace to have a way to use easily the local gitgraph-core
  - [x] Clean up the main package.json and the legacy code
  - [x] Add a storybook to have a great DX / doc
- [x] Deal with the render children props pattern
- [x] Render commits (and make sure that every styles are working as expected)
- [x] Render branches (and make sure that every styles are working as expected)
- [x] Render refs
  - [x] Branches names
  - [x] Tags
- [x] Render commit detail (message, hash…)
[...]

PR #188, opened 2018-01-22
GitGraph.js v2-alpha

nicoespeon:
See #187 for discussions.

Here we can discuss about the specific code changes.


packages/gitgraph-core/src/utils.ts
```diff
@@ -111,16 +111,13 @@ export function toSvgPath(
               points.length > 1 &&
               (i === 1 || i === points.length - 1)
             ) {
+              const previous = points[i - 1];
```
fabien0102: 👍  Good idea!

packages/gitgraph-core/src/commit.ts
```diff
@@ -0,0 +1,174 @@
+import { CommitStyle } from "./template";
+import Branch from "./branch";
+
+export interface CommitOptions {
+  author: string;
+  subject: string;
+  style: CommitStyle;
+  body?: string;
+  notes?: string;
+  refs?: string[];
```
nicoespeon: I wonder if that wouldn't be better not to be optional. We could default it to an empty array. Having to deal with `undefined` value is a bit painful.

Same for `parents`.

What do you think?

packages/gitgraph-core/src/commit.ts
```diff
@@ -0,0 +1,174 @@
...
+   */
+  public style: CommitStyle;
+  /**
+   * Inner text
+   */
+  public innerText?: string;
+  /**
+   * List of branches attached (injected by Gitgraph.withBranches)
+   */
+  public branches?: Array<Branch["name"]>;
```
nicoespeon: I wonder if that wouldn't be better not to be optional. Having to deal with undefined value is a bit painful. 

Plus, `Gitgraph.withBranches()` injects it so technically, at runtime, when we get rendered data this is never `undefined`.

We could default it to an empty array. What do you think?

P.S. same question for `tags`.

nicoespeon: Another discussion about that is: could we have actual `Branch` references instead of just `Branch["name"]`?

Having `Branch`, we could reach for the `name` easily. Having `Branch["name"]` makes us having to find the reference from some map at our disposal, which is more painful. Also, I feel we mostly want to retrieve the actual branch when we get to this point.

But I may be missing something here. What do you think?

P.S. same question for `parents` (having the `Commit` instead of `Commit["hash"]`

fabien0102: I'm totally agree with this, the main challenge is to find a good name for commit in each state (eg: `RawCommit` `CommitWithBranch`…).

Ideally I want to have strong type on all the process, without using `as` or `!` to hack typescript ;)

fabien0102: The initial idea was to mimic the output of `git2json`, since we will probably add a real `gitgraph.import` method, I think that we can refactor this part to simplify everything ;)

Another idea is to link branch and commit with `Map` like our tags pattern. Every pattern works here, we just need to choose "the best one" ^^

packages/gitgraph-core/src/commit.ts
nicoespeon: It makes me wonder about tags. Technically, I understand tags as refs. Branches also are refs. And `HEAD` is another kind of refs. 

But I wouldn't make a difference between the list of `refs` (currently, branch names & HEAD) and `tags`. 

What do you think about merging them?

Though, I agree to expose a different `.tag()` command. However, a `refs` options doesn't make sense to me (it's technical details, shouldn't be part of the public API).

fabien0102: We need a way to show (or not) the refs on the graph, it's why this `refs` is public, for me it's interesting to have different style between `refs` and `branches` and I think that it was the original design for this.

Also remember here that we don't manipulate the `commit` class directly, this class is always instanciate by a `branch` ;) (just to have a little bit more context)

nicoespeon: I'm fine keeping `refs` and `tags` separated for now as I don't have concrete issue because of that 👍

packages/gitgraph-core/src/template.ts
```diff
@@ -0,0 +1,400 @@
+import * as merge from "lodash.merge";
```
mradcliffe: This fails on Typescript 2.8.3 for me (after I manually installed `@types/lodash.merge`

```
src/template.ts:1:24 - error TS2497: Module '"/path/to/gitgraph.js/packages/gitgraph-core/node_modules/@types/lodash.merge/index"' resolves to a non-module entity and cannot be imported using this construct.
```

packages/gitgraph-core/package.json
```diff
@@ -0,0 +1,75 @@
...
+    "build:minify": "uglifyjs -c -m -o ./lib/bundle.min.js -- ./lib/bundle.js"
+  },
+  "devDependencies": {
+    "@types/jest": "^21.1.8",
+    "@types/node": "^9.4.6",
+    "browserify": "^14.5.0",
+    "commitizen": "^2.9.6",
+    "cz-conventional-changelog": "^2.1.0",
+    "jest": "^21.2.1",
+    "lodash.merge": "4.6.1",
```
mradcliffe:
```
$ tsc 
src/template.ts:1:24 - error TS2307: Cannot find module 'lodash.merge'.
```

Probably needs `@types/lodash.merge`?

nicoespeon: Good catch, we added it in 051aa81 👍

packages/gitgraph-node/src/index.ts
```diff
@@ -0,0 +1,21 @@
...
+import computeGraphMap from "./compute-graph-map";
+
+// Infrastructure (logger implementations)
+import consoleGraphLogger from "./console-graph-logger";
+
+function createGitgraph(): GitgraphCore {
+  // Limiting Gitgraph Core config options is intentional since most of them
+  // are not yet implemented in node.js.
+  // Templates options should be limited & customized for the console.
+  return new GitgraphCore();
```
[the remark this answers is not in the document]
nicoespeon: Fair, it went a bit too functional here. +1 for consistency.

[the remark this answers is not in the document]
nicoespeon: So I wonder: do we want to keep `render(gitgraph: Gitgraph)` or should it be `gitgraph.render()` since we use classes here?

I'd say the second one feels more intuitive using classes. But I do also like the render function. What do you think @fabien0102 ?

packages/gitgraph-core/src/gitgraph.ts
```diff
@@ -289,29 +291,27 @@ export class GitgraphCore {
...
    *
    * @param data JSON from `git2json` output
    */
   public import(data: any) {
-    // TODO: validate data format with https://github.com/hapijs/joi
-    this.clear();
+    // Validate `data` format.
+    const schema = Joi.array().items(
+      Joi.object({
+        refs: Joi.array(),
```
fabien0102: Nice 👍 We also need to specify all others params of commit by it's a good start :)

nicoespeon: I suggest we only validate what we actually use, what do you think?

I set `allowUnknown: true` so it won't fail if there are attributes that are not restricted here. This way, it won't break for eventual properties that won't be used here.

fabien0102: Sure, but we are spreading `commit` just after, so we need to have the correct signature ;) (we are actually using almost all fields except `tree` in my memory)

packages/gitgraph-react/src/Tooltip.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  public render() {
+    const commitSize = this.props.commit.style.dot.size * 2;
+    const offset = 10;
+    const padding = 10;
+    const radius = 5;
+    const boxHeight = 50;
+    const boxWidth = offset + this.state.textWidth + 2 * padding;
+
+    const path = [
```
fabien0102: 😍 @nicoespeon Very nice refactor! So clean for a SVG path!

packages/gitgraph-react/src/Gitgraph.tsx
```diff
@@ -210,5 +211,32 @@ export class Gitgraph extends React.Component<GitgraphProps, GitgraphState> {
   }
 }
 
+// For now, this piece of logic is here.
+// But it might be relevant to move this back to gitgraph-core.
+// Ideally, it would be a method of Commit: `commit.message()`.
```
fabien0102: 👍 Totally agree with you about this comment, we should move this into gitgraph-core under `commit.message()`

packages/gitgraph-core/src/commit.ts
```diff
@@ -29,7 +29,7 @@ export interface CommitOptions<TNode> extends CommitRenderOptions<TNode> {
  *
  * @return {string} hex string with 40 chars
  */
-const getRandomHash = () =>
+const getRandomHash = (): string =>
```
fabien0102: 🤔 

image

Typescript already infer this kind of return, so I'm curious about the reason behind this 😄

nicoespeon: Just to make it explicit. But yeah, there was no problem with TS.

mradcliffe:
Confirming output of multimerges look good to me. :tada: Nice work!

nicoespeon:
We're ready for alpha. I updated the repo accordingly, creating `v1` branch, dropping `develop` (we're moving to GitHub flow) and setting `master` as default.

Let's merge this!

party

PR #192, opened 2018-03-17
[2.0.0] gitgraph-react (poc)

fabien0102:
Just for review and talk :wink:

To test:
- pull/checkout to the branch (captain obvious)
- `yarn`
- `cd packages/gitgraph-core && yarn storybook`

Preview: 
gitgraph-react-poc


lerna.json
```diff
@@ -3,5 +3,7 @@
   "packages": [
     "packages/*"
   ],
-  "version": "0.0.0"
+  "version": "2.0.0-alpha",
+  "npmClient": "yarn",
+  "useWorkspaces": true
```
fabien0102: This is to avoid dependencies duplication. https://yarnpkg.com/lang/en/docs/workspaces/

packages/gitgraph-core/package.json
```diff
@@ -71,7 +71,7 @@
   },
   "jest": {
     "transform": {
-      ".(ts)": "<rootDir>/node_modules/ts-jest/preprocessor.js"
+      ".(ts)": "<rootDir>/../../node_modules/ts-jest/preprocessor.js"
```
fabien0102: `ts-jest` is now shared, so on `/node_modules` instead of `/packages/gitgraph-core/node_modules` (yarn workspaces)

packages/gitgraph-core/src/branch.ts
```diff
@@ -74,10 +74,11 @@ export class Branch {
       options.parents.unshift(this.parentCommit.hash);
     }
 
+    const { tag, ...commitOptions } = options;
```
fabien0102: Just realized that we have forgot the syntax `gitgraph.commit({tag: "1.0.0"})` :wink:

packages/gitgraph-react/src/Gitgraph.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from "react";
+import { GitGraphOptions, Commit } from "gitgraph-core/lib/index";
```
fabien0102: My biggest hack for now… I don't why, but it's not works only with `gitgraph-core`… If someone have an idea (normally the package.json is well configure, but I probably missed something)

packages/gitgraph-react/src/GitgraphReact.ts
```diff
@@ -0,0 +1,22 @@
+import { Component } from "react";
+import { GitGraph as GitGraphCore, GitGraphOptions } from "gitgraph-core/lib/index";
```
fabien0102: I don't yet if it's a good idea, but I was thinking about renaming `GitGraph` to `GitgraphCore` to avoid this kind of pattern and also permit to consolidate every types (I want to be able to load `Branch`, `Commit`… from `gitgraph-react` directly)

packages/gitgraph-react/src/stories/index.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  public handleChange = (name: string) => (e: React.SyntheticEvent<HTMLInputElement>): void => {
+    this.setState({ [name]: e.currentTarget.value });
+  }
+
+  public render() {
+    return (
+      <div>
+        <form onSubmit={(e) => e.preventDefault() || this.addCommit()}>
```
fabien0102: For now it's light and ugly, but I will maybe add semantic-ui to have a better playground

fabien0102:
mmm, we need to update the travis config… this project have a hard dependency to yarn for the workspaces feature

dolanmiu:
https://github.com/nicoespeon/gitgraph.js/blob/2.0.0-gitgraph-react/packages/gitgraph-react/src/Gitgraph.tsx#L25-L48

I think we should put all the drawing of all `svg` and `circles` in another package, otherwise `gitgraph-angular` would need to repeat the drawing!

fabien0102:
@dolanmiu Totally agree to avoid any duplication, for instant I'm just trying to see if everything is correct for the core implementation (and it's not correct yet ^^) I'm currently adding the "branches paths calculation" into the core (and will probably break everything with the refactor of `gitgrah.log`, `gitgraph.render` to add directly in the constructor something like `onUpdate`, so no more needs to extends `gitgraph-core` for any implementation :wink:

For the SVG part, I really want to be able to animate any graph, and add the availability to override any `commit`. In react something like this:
```js
const MyGraph = () => (
  <Gitgraph>
    <Branch name="master" />
    <Commit render={MyCustomCommit} />
    {/* ... */}
  </Gitgraph>
)
```
For instant it's just a draft of the API of course, but just to keep this in mind. I don't really know your needs for the angular spec (my last angular xp was with the 2.0.0) so I think that I will just POC on react until stabilize the core API, and after we can think about extracting any logics that can be share with other libs :wink:

Our goal is of course to avoid any useless duplication, but also to have the more flexibility and maintability as possible 😃

nicoespeon:
As discussed IRL with @fabien0102: we're merging that into the `2.0.0` as they are a lot of work on the core in this PR. This branch has no much value in itself.

Also, we'll fix Travis tests in there too.

PR #196, opened 2018-07-05
Add the ability to put some text in the commit bubble. (commitDotText)

carlorivera:
Add ability to put some text in the commit bubble. 
image

I'm using this to generate some docs and find it useful for referring to commits.

### DIFF
--- a/examples/index.js
+++ b/examples/index.js
@@ -77,6 +77,7 @@ var commitConfig = {
   dotStrokeWidth: 10,
   messageHashDisplay: false,
   messageAuthorDisplay: true,
+  messageShort: "C1",
   message: "Alors c'est qui le papa ?",
   tooltipDisplay: false,
   author: "Me <me@planee.fr>"
@@ -128,7 +129,8 @@ test.merge(master, "My special merge commit message");
 
 // Then, continue committing on the "test" branch
 test.commit({
-  message: "It works !"
+  message: "It works !",
+  messageShort: "C2"
 });
 
 var fastForwardBranch = test.branch("fast-forward");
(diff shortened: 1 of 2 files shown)


src/gitgraph.js
```diff
@@ -1053,6 +1053,7 @@
    * @param {number[]} [options.lineDash = this.template.commit.dot.lineDash]
    *
    * @param {string} [options.message = "He doesn't like George Michael! Boooo!"] - Commit message
+   * @param {string} [options.message] - short commit message (A few chars) to appear on the commit dot
```
fabien0102: Bad copy/past here, this should be `[options.messageShort]` :wink:

src/gitgraph.js
```diff
@@ -1205,6 +1207,19 @@
       }
     }
 
+    if (this.messageShort !== undefined) {
```
fabien0102: You can write `if (this.messageShort)` here, to be consistant with the actual code style.

examples/index.js
```diff
@@ -77,6 +77,7 @@ var commitConfig = {
   dotStrokeWidth: 10,
   messageHashDisplay: false,
   messageAuthorDisplay: true,
+  messageShort: "C1",
```
nicoespeon: I don't think `messageShort` is explicit enough for people without context.

What do you think about something like `commitDotText`? Something else?

carlorivera: Great idea. I'll update.

nicoespeon:
Good for me too. Merging it.
Thanks @carlorivera 👍

PR #197, opened 2018-07-07
Fix Prettier config

nicoespeon:
- [x] Move Prettier configuration up to the root directory
- [x] Configure Prettier with config files
- [x] Configure files we don't want to prettify (ignored)
- [x] Fix pre-commit hook, working across all packages
- [x] Run Prettier once over everything


nicoespeon:
@fabien0102 I made a PR of it so you can review and tell me if that looks good to you.

As far as I've tested, it works on pre-commit hooks, for any file.

Also, I omitted JSON files because I don't like having long strings moved to another line. Also, I think that JSON format is already pretty standard and there is no much to prettify here. 

What do you think?

package.json
```diff
@@ -20,7 +20,10 @@
     "javascript"
   ],
   "devDependencies": {
-    "lerna": "^2.5.1"
+    "husky": "1.0.0-rc.12",
```
nicoespeon: The good part with `husky@next` is that it knows when you commit a partially staged file. 

Used with `pretty-quick`, if the file is modified during precommit hook (because Prettier), then it'll warn you before adding the whole file again. 

So you can partially stage files without having precommit hook adding the whole stuff on fixes 🎉 

image

fabien0102: My hero! :tada: This look very amazing!

nicoespeon:
@fabien0102 THX! 😄 

I'll look at tslint in a separate issue (adding it to the main checklist). It may be worth moving the config up to the root directory for that too. Same goes for commitizen & jest.

PR #198, opened 2018-07-07
Gitgraph node

nicoespeon:
Gitgraph rendering implementation, in node.js. So you can render graph in your terminal too 😎


nicoespeon:
@fabien0102 it would be awesome if you could review the code to give me some feedback before I merge that into v2 branch.

Also, it'd be even more awesome if you could explain me what I did wrong with Travis here. It seems it doesn't find `gitgraph-core/lib` and I don't understand why… What's different from `gitgraph-react`?

fabien0102:
For Travis, you may add a build gitgraph-core before maybe, we don't have any unit test in gitgraph-react (only storybook) so we have probably forgot this build step ;) (but I need to check, I'm on my phone now ^^)

packages/gitgraph-node/src/compute-graph-map.test.ts
```diff
@@ -0,0 +1,329 @@
+import { GitgraphCore, Branch, Template } from "gitgraph-core/lib";
```
fabien0102: Why do you need to import gitgraph-core here? We are testing gitgraph-node here so we should only import from gitgraph-node

packages/gitgraph-node/src/index.ts
```diff
@@ -0,0 +1,17 @@
+import { GitgraphCore } from "gitgraph-core/lib";
+
+// Domain (rendering logic)
+import computeGraphMap from "./compute-graph-map";
+
+// Infrastructure (logger implementations)
+import consoleGraphLogger from "./console-graph-logger";
+
+const gitgraph = new GitgraphCore();
```
fabien0102: With this pattern, we can't personalize the template (eg: colors in this particular usecase) I will prefer to give a gitgraph instance to render to avoid this lack of config

packages/gitgraph-node/src/index.ts
```diff
@@ -0,0 +1,17 @@
...
+
+function render() {
+  const graphMap = computeGraphMap(gitgraph);
+  consoleGraphLogger(graphMap);
+}
+
+export { gitgraph, render };
+export * from "gitgraph-core/lib";
```
fabien0102: Because all template values are not relevant here, we can maybe limit the gitgraph options ;)

packages/gitgraph-node/src/compute-graph-map.test.ts
fabien0102: BTW, we should be able to use `gitgraph-core` directly (without the `/lib`) I know that I've done the same hack into gitgraph-react but it's a hack ^^

nicoespeon: For the record:
- we didn't manage to get rid of `/lib`, yet
- yes, we are testing `gitgraph-node` here, but we need to use types defined in `gitgraph-core/lib` and we set up some custom template to test `computeGraphMap(gitgraph)`

PR #201, opened 2018-08-19
Add dotFont option and a default for it.

prudy:
If the dot text is used, it is not drawn with the correct value. The first commit uses some defaults, then font changes according to message font.

This patch solves it, adding config option 'font' and 'dotFont' for commit/merge.

### DIFF
--- a/src/gitgraph.js
+++ b/src/gitgraph.js
@@ -626,6 +626,7 @@
    * @property {string} [tagFont = this.template.commit.tag.font] - Font of the tag
    * @property {string} [displayTagBox = true] - If true, display a box around the tag
    *
+   * @property {string} [fotFont = this.template.commit.dot.font] - Font of the dot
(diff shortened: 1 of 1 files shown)


src/gitgraph.js
```diff
@@ -626,6 +626,7 @@
    * @property {string} [tagFont = this.template.commit.tag.font] - Font of the tag
    * @property {string} [displayTagBox = true] - If true, display a box around the tag
    *
+   * @property {string} [fotFont = this.template.commit.dot.font] - Font of the dot
```
nicoespeon: A little typo here: *dotFont

prudy:
Done.

nicoespeon:
Thanks @prudy. I published your change under [v1.13.0](https://github.com/nicoespeon/gitgraph.js/releases/tag/v1.13.0) 👍

PR #214, opened 2018-12-19
Add commit object to the mouseEventOptions object

srangos:
This will allow users to work with the full commit object during mouse over and mouse out events


nicoespeon:
That's a good idea. Thanks @srangos 👍

PR #216, opened 2019-01-24
fix arrow color problem 

yappynoppy:
The color of some arrow was not the same color of stroke with it.
[before]
Imgur

So, this problem will be fixed.
[after]
Imgur


nicoespeon:
Thank you for this fix 👍

PR #230, opened 2019-03-05
Fix/blackarrow arrow orientation

### LINKED ISSUE #226: Fix blackarrow arrow orientation
See branch `fix/blackarrow-arrow-orientation`. I updated **blackArrow** story to illustrate the problem.

image

nicoespeon:
Fixes #226

Result:

image

### DIFF
--- a/packages/gitgraph-core/src/utils.ts
+++ b/packages/gitgraph-core/src/utils.ts
@@ -160,7 +160,7 @@ function getAlpha<TNode = SVGElement>(
   let alphaY;
   let alphaX;
 
-  // Angle always start from previous commit Y position:
+  // Angle usually start from previous commit Y position:
   //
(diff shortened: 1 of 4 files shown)


packages/gitgraph-core/src/utils.ts
```diff
@@ -193,26 +194,24 @@ function getAlpha<TNode = SVGElement>(
       break;
   }
 
+  // If commit is distant from its parent, there should be no angle.
```
nicoespeon: At first, I thought this applied only to reverse arrow scenarios. But it turns out I was wrong. Now it works for every scenario.

nicoespeon: cc @fabien0102 ;-)

PR #231, opened 2019-03-05
React/branches labels

### LINKED ISSUE #142: Mode: Branch pointers only

### LINKED ISSUE #160: Branch labels are shown on first commit in branch, not head of branch

### LINKED ISSUE #221: Render branch labels out of commit message
- [x] Render branch labels as tags
- [x] Fix fast-forward merge scenario
image

- [x] Render branch labels in horizontal mode
- [x] Re-introduce branch label style options (stroke color, bg color, font) (template, graph & branch level)
- [ ] Replace commit `displayBranch` with a `showLabel` equivalent options on branch instead

nicoespeon:
Fixes #221 

Also, I realized that it will:
- Fix #160 (branch names display on the last commit of the branch) 
- Fix #142 (branch names display in compact mode, pointing to the commit)

## Default

image

## Vertical reverse

image

## Horizontal

image

## Horizontal reverse

image

## Blackarrow

image

## Custom

image

### DIFF
--- a/packages/gitgraph-react/src/Gitgraph.tsx
+++ b/packages/gitgraph-react/src/Gitgraph.tsx
@@ -127,14 +127,11 @@ class Gitgraph extends React.Component<GitgraphProps, GitgraphState> {
       );
     }
 
-    if (this.$commits.current) {
-      this.translateCommitMessages(Array.from(this.$commits.current.children));
-    }
-
     if (!this.state.shouldRecomputeOffsets) return;
     if (!this.$commits.current) return;
 
     const commits = this.$commits.current.children;
+    this.translateCommitMessages(Array.from(commits));
     this.setState({
       commitYWithOffsets: this.computeOffsets(commits),
       shouldRecomputeOffsets: false,
--- a/packages/gitgraph-react/src/Gitgraph.tsx
+++ b/packages/gitgraph-react/src/Gitgraph.tsx
@@ -162,37 +162,34 @@ class Gitgraph extends React.Component<GitgraphProps, GitgraphState> {
 
   private renderBranchesLabels() {
     const branches = Array.from(this.gitgraph.branches.values());
-    return branches
-      .map((branch) => {
-        const commitHash = this.gitgraph.refs.getCommit(branch.name);
-        return this.state.commits.find(({ hash }) => commitHash === hash);
-      })
-      .filter(
-        (commit): commit is Commit<ReactSvgElement> => commit !== undefined,
-      )
-      .filter((commit) => commit.style.message.displayBranch)
-      .map((commit) => {
-        const ref = React.createRef<SVGGElement>();
+    return branches.map((branch) => {
+      const commitHash = this.gitgraph.refs.getCommit(branch.name);
+      const commit = this.state.commits.find(({ hash }) => commitHash === hash);
 
-        // Store the ref to adapt commit message position.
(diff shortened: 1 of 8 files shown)


packages/gitgraph-react/src/Gitgraph.tsx
```diff
@@ -223,7 +294,11 @@ class Gitgraph extends React.Component<GitgraphProps, GitgraphState> {
 
   private renderMessage(commit: Commit<ReactSvgElement>) {
     if (commit.renderMessage) {
-      return commit.renderMessage(commit, this.state.commitMessagesX);
+      return (
+        <g className="message" transform="translate(0, 0)">
```
nicoespeon: Explain translate(0, 0)

packages/gitgraph-react/src/Gitgraph.tsx
```diff
@@ -118,10 +123,36 @@ class Gitgraph extends React.Component<GitgraphProps, GitgraphState> {
         "height",
         // As we translate the graph by `Tooltip.padding`,
         // we need to take it into account in height calculation too.
-        (height + Tooltip.padding).toString(),
+        (height + Tooltip.padding + BranchLabel.paddingX).toString(),
       );
     }
 
+    if (this.$commits.current) {
+      Array.from(this.$commits.current.children).forEach((commit) => {
```
nicoespeon: Extract and name operation

packages/gitgraph-react/src/Gitgraph.tsx
```diff
@@ -118,10 +123,36 @@ class Gitgraph extends React.Component<GitgraphProps, GitgraphState> {
         "height",
         // As we translate the graph by `Tooltip.padding`,
         // we need to take it into account in height calculation too.
-        (height + Tooltip.padding).toString(),
+        (height + Tooltip.padding + BranchLabel.paddingX).toString(),
```
nicoespeon: Explain

packages/gitgraph-react/src/Gitgraph.tsx
nicoespeon: Actually useless. WIP that was committed. I removed it.

awgeorge:
It maybe nice to have the option to change where the label sits. As then I could move the overlap name to the first commit.

Screenshot 2019-08-06 at 15 30 19

nicoespeon:
@awgeorge you're right, that would probably help in 90% of the case.

To fix this behaviour completely, I think we could re-think how we display branch labels in horizontal mode. Maybe we could put the labels on the left of the graph in this scenario:

image

As of today, the workaround is not super accessible, but still doable. You can technically control where to put the branch labels with a custom renderer, see https://github.com/nicoespeon/gitgraph.js/issues/319#issuecomment-517998036

That being said, I'd accept a PR to introduce an option to put the branch label at the first commit of the branch, to cover the 90% of cases 😄

PR #234, opened 2019-03-07
Fix and improve the `translateCommitMessagesWithBranchLabel` strategy

fabien0102:
In `translateCommitMessagesWithBranchLabel`, `message.getAttribute("transform")` can be null in custom message render cases. This PR fix these edge cases.

## 6. Custom Render / with render message (after fix)

image

### DIFF
--- a/packages/gitgraph-react/src/Gitgraph.tsx
+++ b/packages/gitgraph-react/src/Gitgraph.tsx
@@ -356,18 +356,21 @@ class Gitgraph extends React.Component<GitgraphProps, GitgraphState> {
       const message = commit.getElementsByClassName("message")[0];
       if (!message) return;
 
-      const matches = message
-        .getAttribute("transform")!
-        .match(/translate\((\d+),\s(\d+)\)/);
+      const transformAttribute =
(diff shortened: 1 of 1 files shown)


packages/gitgraph-react/src/Gitgraph.tsx
```diff
@@ -356,18 +356,21 @@ class Gitgraph extends React.Component<GitgraphProps, GitgraphState> {
...
+      const [, x] = matches.map((a) => parseInt(a, 10));
       // For some reason, 1 padding is not included in BBox total width.
       const branchLabelWidth =
         branchLabel.current.getBBox().width + BranchLabel.paddingX;
       const padding = 10;
       const newX = x + branchLabelWidth + padding;
-      message.setAttribute("transform", `translate(${newX}, ${y})`);
+      message.setAttribute(
+        "transform",
+        transformAttribute.replace(/translate\((\d+),/, `translate(${newX},`),
```
fabien0102: This line should avoid future bugs with custom `transform` attribute 😉 We don't erase any initial value anymore (eg: scale, rotation…)

PR #239, opened 2019-03-21
Customize Tag style

### LINKED ISSUE #151: Customize tag text color + font
Branch colors that look good on white backgrounds tend to be darker colors. Unfortunately, it seems the tag text color is always black, so it can become hard to read the tag colors. Optimally, there would be two sets of colors--perhaps something like this:
```
var template = GitGraph.prototype.newTemplate("metro");
        
template.branchColors = [ "FireBrick", "SeaGreen", "DodgerBlue", "DarkViolet" ];
template.tagColors = [ "White", "White", "White", "White" ];
```

nicoespeon:
Fixes #151

Tags can be customized at template & tag level (through both gitgraph and branch user APIs):

image

Default style for blackarrow:

image

### DIFF
--- a/packages/gitgraph-core/src/tag.ts
+++ b/packages/gitgraph-core/src/tag.ts
@@ -1,0 +1,37 @@
+import { TagStyle, CommitStyle, DEFAULT_FONT } from "./template";
+import { numberOptionOr } from "./utils";
+
+export { Tag };
+
+class Tag {
+  /**
+   * Name
+   */
+  public readonly name: string;
+  /**
+   * Style
+   */
+  public get style(): TagStyle {
+    return {
+      strokeColor: this.tagStyle.strokeColor || this.commitStyle.color,
+      bgColor: this.tagStyle.bgColor || this.commitStyle.color,
+      color: this.tagStyle.color || "white",
+      font: this.tagStyle.font || this.commitStyle.message.font || DEFAULT_FONT,
+      borderRadius: numberOptionOr(this.tagStyle.borderRadius, 10),
+      pointerWidth: numberOptionOr(this.tagStyle.pointerWidth, 12),
+    };
+  }
+
+  private readonly tagStyle: Partial<TagStyle>;
+  private readonly commitStyle: CommitStyle;
+
+  constructor(
+    name: string,
+    tagStyle: Partial<TagStyle>,
+    commitStyle: CommitStyle,
+  ) {
+    this.name = name;
+    this.tagStyle = tagStyle;
+    this.commitStyle = commitStyle;
+  }
+}
--- a/packages/gitgraph-core/src/user-api/gitgraph-user-api.ts
+++ b/packages/gitgraph-core/src/user-api/gitgraph-user-api.ts
@@ -27,9 +27,10 @@ interface GitgraphCommitOptions<TNode> extends CommitRenderOptions<TNode> {
   onMouseOut?: (commit: Commit<TNode>) => void;
 }
 
-interface GitgraphTagOptions<TNode> {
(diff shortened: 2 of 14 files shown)


packages/gitgraph-core/src/user-api/gitgraph-user-api.ts
```diff
@@ -99,30 +105,43 @@ class GitgraphUserApi<TNode> {
...
+    }
+
     if (!ref) {
       const head = this._graph.refs.getCommit("HEAD");
       if (!head) return this;
 
       ref = head;
     }
 
-    if (typeof ref !== "string") {
```
nicoespeon: This case was impossible: end-users don't have access to `Commit` objects.

packages/gitgraph-core/src/user-api/gitgraph-user-api.ts
```diff
@@ -99,30 +105,43 @@ class GitgraphUserApi<TNode> {
...
    * Tag a specific commit.
    *
    * @param name Name of the tag
    * @param ref Commit or branch name or commit hash
    */
   public tag(
-    name: string,
-    ref?: Commit<TNode> | Commit["hash"] | Branch["name"],
-  ): this {
+    name: GitgraphTagOptions["name"],
```
nicoespeon: Avoid duplicating the type between all signatures.

think-about-it

packages/gitgraph-core/src/user-api/gitgraph-user-api.ts
```diff
@@ -139,6 +158,7 @@ class GitgraphUserApi<TNode> {
     }
 
     this._graph.tags.set(name, commitHash);
+    this._graph.tagStyles[name] = style;
```
nicoespeon: Simplest solution I found was to store the style in a map, then to retrieve it when we compute commit style (tag uses commit style as fallback).

packages/gitgraph-react/src/Tag.tsx
```diff
@@ -1,17 +1,16 @@
...
+  tag: Tag;
 }
 
 interface State {
   textWidth: number;
   textHeight: number;
 }
 
-export class Tag extends React.Component<Props, State> {
+export class TagElement extends React.Component<Props, State> {
```
nicoespeon: Necessary rename to avoid conflict with core `Tag` element (could have renamed the import, but I found `TagElement` to be a better option).

packages/gitgraph-react/src/Gitgraph.tsx
```diff
@@ -351,8 +351,12 @@ class Gitgraph extends React.Component<GitgraphProps, GitgraphState> {
       const ref = this.createTagRef(commit);
 
       return (
-        <g key={`${commit.hashAbbrev}-${tag}`} ref={ref}>
-          <Tag name={tag} commit={commit} />
+        <g
+          key={`${commit.hashAbbrev}-${tag.name}`}
+          ref={ref}
+          transform={`translate(0, ${commit.style.dot.size})`}
```
nicoespeon: Moved transform here so `<Tag>` doesn't have to know about the commit stuff.

packages/gitgraph-core/src/tag.ts
```diff
@@ -0,0 +1,37 @@
...
+
+class Tag {
+  /**
+   * Name
+   */
+  public readonly name: string;
+  /**
+   * Style
+   */
+  public get style(): TagStyle {
```
nicoespeon: Using a getter to compute the `tag.style` at runtime. I find it a very nice alternative to computing everything in the constructor.

Issue #243, opened 2019-03-26
Enforce branch column order and colors


haizz:
First of all, thank you for the great project!

Currently the order of branch columns is determined by which branch comes first in commit list:

```typescript
export class GraphColumns<TNode> {
  private branches: Set<Branch["name"]> = new Set();

  public constructor(commits: Array<Commit<TNode>>) {
    commits.forEach((commit) => this.branches.add(commit.branchToDisplay));
  }
// ...
}
```

My app has an UI which hides/shows branches based on user requests by filtering commits which get imported to gitgraph, and arbitrarily changing branch order and colors become confusing for users. I'd like to have master branch always having column index 0, first selected branch having column index 1, second branch having column index 2, ... regardless of which commit comes first in the log. 

What would be the easiest way to accomplish this?

haizz:
BTW, I'm using `import` function from `GitgraphUserApi`.

fabien0102:
You can specify the color in "manual" mode with `gitgraph.branch({name: "mybranch", style: {color: "pink"}})` -> https://codesandbox.io/embed/ym2jrq408j (and the dot color is not following ^^ time to open a new issue 😅 )

For the `import`, you don't have (yet) any option to customize this, but it's really interesting problem ^^
I guess we can add a `branchToColor` options somewhere for this kind of case.

```tsx
<Gitgraph options={{branchToColor: (branch) => "black"}}></Gitgraph>
```

haizz:
Will that enforce the column order? What if I would like to keep recent branches closer to the master? Currently it's strictly earlier commits who receive lower column index.

fabien0102:
mmmm, you're right, maybe a `sortBranch` 🤔 with the same API as the native `Array.sort`

haizz:
I can make a PR if you define an API and throw me some basic guidelines on where/how to implement this.

haizz:
Maybe just pass an array of branch names with desired order?

fabien0102:
Sure but in the case of the `import` how do you know what branch name you have? same in the classic mode, the config is the first thing you give, so seams a bit fragile to give an array of branch names.

fabien0102:
I don't exactly remember how where is defined the branch order, I need to dig a bit in the code 😅 

As guideline, all the project is tested (unit tests with jest for the `core`, visual tested with storybook for `react` and `js`), so depending where is the problem, we always begin to add a test or a story to demonstrate the bug, and after fix it! (TDD style ^^)

[...]

PR #246, opened 2019-03-28
Refactor branches order

nicoespeon:
Extract the concept of "BranchesOrder" (formerly "GraphColumns"), following this discussion: https://github.com/nicoespeon/gitgraph.js/issues/243#issuecomment-476955667

Also refactor the logic of commit message into @gitgraph/core.


nicoespeon:
@fabien0102 it's so good to have Chromatic automatically check there is no regression on the rendering lib. Thanks for pushing it 👍

PR #247, opened 2019-03-28
Add BranchesOrderFunction option to set user-defined branches order

### LINKED ISSUE #243: Enforce branch column order and colors

haizz:
Fixes #243

### DIFF
--- a/packages/gitgraph-core/src/__tests__/gitgraph.getRenderedData.branchesPaths.test.ts
+++ b/packages/gitgraph-core/src/__tests__/gitgraph.getRenderedData.branchesPaths.test.ts
@@ -1,5 +1,6 @@
 import { GitgraphCore } from "../gitgraph";
 import { Orientation } from "../orientation";
+import { BranchesOrderFunction } from "../branch";
 import { metroTemplate, TemplateName } from "../template";
 
 describe("Gitgraph.getRenderedData.branchesPaths", () => {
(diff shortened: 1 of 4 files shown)


packages/gitgraph-core/src/__tests__/gitgraph.getRenderedData.branchesPaths.test.ts
```diff
@@ -438,4 +439,49 @@ describe("Gitgraph.getRenderedData.branchesPaths", () => {
     expect(result[1][0].name).toBe("feat");
     expect(result[1][0].computedColor).toBe("#000000");
   });
+
+  it("should generate branches paths if I define branches order function", () => {
```
nicoespeon: 👍 for adding a test to illustrate the behavior!

I think we could simplify the test a bit, just to focus on the important part. But no worries, I'll refactor the code a bit myself after merging, there is no big deal ;)

packages/gitgraph-core/src/branch.ts
```diff
@@ -44,6 +45,17 @@ interface BranchOptions<TNode = SVGElement> {
   onGraphUpdate: () => void;
 }
 
+/**
+ * Type of a function which defines branch order in computed branches paths.
+ * Returns:
+ * - negative number if `branchNameA` should be rendered before `branchNameB`
+ * - positive number if `branchNameA` should be rendered after `branchNameB`
+ */
+type BranchesOrderFunction = (
```
nicoespeon: I'll move this type into the `BranchesOrder` file as these concepts feels definitely related. It's subtly different from `Branch`, which represent only 1 branch where `BranchesOrder` represent the ordering of all branches in the graph.

I liked that you wrote a comprehensive doc for this type!

nicoespeon: I might rename this `SortBranches` though. The fact it's a function is already embed in the type. I like to make it explicit what this function is about 👌

packages/gitgraph-core/src/gitgraph.ts
```diff
@@ -38,6 +39,7 @@ interface GitgraphOptions {
   author?: string;
   commitMessage?: string;
   generateCommitHash?: () => Commit["hash"];
+  branchesOrderFunction?: BranchesOrderFunction;
```
nicoespeon: As said previously, I'll merge the code since it's good, but I'll probably rename this `sortBranches` before publishing.

PR #253, opened 2019-04-02
Factor expensive computations out of commit iterations

haizz:
Currently there are some heavy computations on each commit iteration, for example `withBranches` repeatedly calls `getBranches` of whole graph for each and every commit. This makes performance drop very fast when number of commits is rising. These whole graph computations can be simply factored out and performed just once before iterating through commits.

Average timings on my data (506 commits) before this PR:

import: 561 ms
getRenderedData: 502 ms

After this PR:

import: 100 ms
getRenderedData: 13 ms

@nicoespeon @fabien0102

### DIFF
--- a/packages/gitgraph-core/src/gitgraph.ts
+++ b/packages/gitgraph-core/src/gitgraph.ts
@@ -6,6 +6,7 @@ import {
 } from "./branch";
 import { Commit } from "./commit";
 import { createGraphRows } from "./graph-rows";
+import { RegularGraphRows } from "./graph-rows/regular";
 import { BranchesOrder, CompareBranchesOrder } from "./branches-order";
 import {
   Template,
@@ -193,21 +194,26 @@ class GitgraphCore<TNode = SVGElement> {
    * Return commits with data for rendering.
    */
   private computeRenderedCommits(): Array<Commit<TNode>> {
+    const branches = this.getBranches();
     const commitsWithBranches = this.commits.map((commit) =>
-      this.withBranches(commit),
+      this.withBranches(branches, commit),
+    );
+
+    const rows = createGraphRows(this.mode, this.commits);
+    const branchesOrder = new BranchesOrder<TNode>(
+      commitsWithBranches,
+      this.template.colors,
+      this.branchesOrderFunction,
     );
 
     return (
       commitsWithBranches
         .map((commit) => commit.setRefs(this.refs))
-        .map((commit) => this.withPosition(commitsWithBranches, commit))
+        .map((commit) => this.withPosition(rows, branchesOrder, commit))
         // Fallback commit computed color on branch color.
         .map((commit) =>
           commit.withDefaultColor(
-            this.getBranchDefaultColor(
-              commitsWithBranches,
-              commit.branchToDisplay,
-            ),
+            this.getBranchDefaultColor(branchesOrder, commit.branchToDisplay),
           ),
         )
         // Tags need commit style to be computed (with default color).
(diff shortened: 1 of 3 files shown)


haizz:
By the way, most of the time (~80%) in `import` is spent for input validation. I don't know, but in my case it's quite unnecessary, because my code which generates this git json is already verified by TypeScript. And 80 ms latency is pretty noticeable in my app which updates and re-imports commits on certain user interactions. But that's probably a topic for discussion in a separate issue.

nicoespeon:
Hi @haizz 👋 

Sorry, I didn't get time to review your PR yet. I'll do it soon.

So far, performance was not a concern indeed. It becomes relevant as `import()` makes it more common to generate large graphs. Thanks for taking a stab at it. I've some stuff in mind that would considerably improve rendering performance indeed.

As for yup, I just removed it in the latest version of the lib: https://github.com/nicoespeon/gitgraph.js/pull/257

It turned out it was only used in `import()` and impacted the bundlesize badly. I added a couple of basic, manual validation. If it still takes too much computation time, we could provide a way to opt-out the validation if you know that you're using `git2json` output anyway—which should be the case for most people.

packages/gitgraph-core/src/gitgraph.ts
```diff
@@ -265,11 +277,13 @@ class GitgraphCore<TNode = SVGElement> {
   /**
    * Add `branches` property to commit.
    *
+   * @param branches All branches mapped by commit hash
    * @param commit Commit
    */
-  private withBranches(commit: Commit<TNode>): Commit<TNode> {
-    const branches = this.getBranches();
-
```
nicoespeon: 👍

packages/gitgraph-core/src/gitgraph.ts
```diff
@@ -317,22 +331,18 @@ class GitgraphCore<TNode = SVGElement> {
...
   ): Commit<TNode> {
-    const rows = createGraphRows(this.mode, this.commits);
     const row = rows.getRowOf(commit.hash);
     const maxRow = rows.getMaxRow();
 
-    const branchesOrder = new BranchesOrder<TNode>(
-      commitsWithBranches,
-      this.template.colors,
-      this.branchesOrderFunction,
-    );
```
nicoespeon: 👍

packages/gitgraph-core/src/gitgraph.ts
```diff
@@ -369,19 +379,13 @@ class GitgraphCore<TNode = SVGElement> {
...
-    commits: Array<Commit<TNode>>,
+    branchesOrder: BranchesOrder<TNode>,
     branchName: Branch["name"],
   ): string {
-    const branchesOrder = new BranchesOrder<TNode>(
-      commits,
-      this.template.colors,
-      this.branchesOrderFunction,
-    );
-
```
nicoespeon: 👍

packages/gitgraph-core/src/graph-rows/regular.ts
```diff
@@ -14,7 +16,10 @@ export class RegularGraphRows<TNode> {
   }
 
   public getMaxRow(): number {
-    return uniq(Array.from(this.rows.values())).length - 1;
+    if (this.maxRowCache === undefined) {
+      this.maxRowCache = uniq(Array.from(this.rows.values())).length - 1;
+    }
+    return this.maxRowCache;
```
nicoespeon: 👍 Thanks for adding cache.

The only thing is that we should invalidate the cache whenever `this.rows` changes. That means we should reset it to `undefined` in the `computeRowsFromCommits()` method. It's fine with current usage as we probably only call the compute method once per instantiation, but it would prevent future caching issues.

I'll do that after merge.

packages/gitgraph-core/src/user-api/gitgraph-user-api.ts
```diff
@@ -227,8 +227,9 @@ class GitgraphUserApi<TNode> {
     });
 
     // Create branches.
+    const branches = this._getBranches();
     this._graph.commits
-      .map((commit) => this._withBranches(commit))
+      .map((commit) => this._withBranches(branches, commit))
```
nicoespeon: 👍

haizz:
@nicoespeon Thanks for having thorough tests :)

nicoespeon:
@haizz FYI, I just did a release of the different packages that includes this change + another perf optimization. 

Let me know if that's better or if it generates issues for you.

Issue #202, opened 2018-08-22
Ensure bundle size is still reasonable


nicoespeon:
Gitgraph.js v1 had no dependency. As we work on v2 (#188), we start using external libs. This may end up with a bigger bundle size at the end, and we want to keep it reasonable for front-end libs.

We should have a look at final bundle size (and optimise it) when v2 is released.

Overall, we can see that `@gitgraph/core` is already 52kB gzipped 😱 
https://bundlephobia.com/result?p=@gitgraph/core

But digging down, we can see it's mostly lodash here (then yup). It's probably not tree-shaked yet, so we should go and optimize this:

![image](https://user-images.githubusercontent.com/1094774/55291166-6cb6ab80-53a9-11e9-8e8c-4001d3bae014.png)

I'd expect the final gzipped bundle to be < 20kB.

As a matter of comparison, `gitgraph.js` (v1) was ~6kB gzipped: https://bundlephobia.com/result?p=gitgraph.js

P.S: A tool like https://github.com/siddharthkp/bundlesize will help keeping a eye on our bundle size. It can be configured as PR check.

nicoespeon:
`@gitgraph/js` is now 9kB gzipped. It's almost what `gitgraph.js` was, and below my initial target of 20kB  🎉 

Considering this as solved.

PR #256, opened 2019-04-07
Drop lodash from @gitgraph/core

nicoespeon:
Related to #202.

It turned out we only needed lodash for 2 functions. But it took 66% of the final `@gitgraph/core` bundle. It doesn't worth it.

I tried to check for tree-shaking, but I couldn't get `lodash-es` to work with the way we bundle the library. It should be compatible with browsers, so I think we're stuck with CommonJS and can't get tree-shaking. 

I might be wrong though, but I didn't want to waste too much time on this and decided to replace the functions we used with alternative implementation instead.

| Before | After |
|---|---|
| image | image |

### DIFF
--- a/packages/stories/package.json
+++ b/packages/stories/package.json
@@ -1,5 +1,6 @@
 {
   "name": "@gitgraph/stories",
+  "version": "0.0.0",
   "private": true,
   "description": "Stories of @gitgraph rendering libraries for E2E tests",
   "author": "Nicolas Carlo <nicolascarlo.espeon@gmail.com>",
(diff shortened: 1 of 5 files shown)


packages/stories/package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "@gitgraph/stories",
+  "version": "0.0.0",
```
nicoespeon: I removed it so lerna doesn't bother updating it. But it turns out package.json is not valid without, thus causing E2E tests to fail as it couldn't install `chromatic` anymore 🤷‍♂️

PR #273, opened 2019-04-14
Multiple branches from same commit

nicoespeon:


joneserick:
@nicoespeon is this PR still valid? I'm really interested in using this feature of multiple branches from the same commit.

nicoespeon:
@joneserick well it is because it only added the tests showing that the feature is missing.

You can see that there's no `develop` branch showing up here: https://www.chromaticqa.com/snapshot?appId=5c9632d9ead30500207bb576&id=5cb3b3a140be8b0020b2216b

image

There should be.

The goal would be to make the tests pass.

packages/stories/src/gitgraph-js/1-basic-usage.stories.tsx
```diff
@@ -226,6 +226,21 @@ storiesOf("gitgraph-js/1. Basic usage", module)
...
   ))
+  .add("multiple branches at the same commit", () => (
+    <GraphContainer>
+      {(graphContainer) => {
+        const gitgraph = createGitgraph(graphContainer, {
+          generateCommitHash: createFixedHashGenerator(),
+        });
+
+        const master = gitgraph.branch("master").commit("one");
+        const feat1 = gitgraph.branch("feat/1");
```
nicoespeon, line 237:
```suggestion
        const feat1 = gitgraph.branch("feat/1-trigger-ci");
```

iliakan:
Will multiple branches show correctly, if the structure is imported via `gitgraph.import()`?

PR #295, opened 2019-05-15
feat: Create branch from hash

kohakukun:
- Accept hash as `from.name` when creating a branch

I have a use case where I want to go back in the commit and branch from that particular hash, is there any other supported why?

### DIFF
--- a/packages/gitgraph-core/src/__tests__/branch.test.ts
+++ b/packages/gitgraph-core/src/__tests__/branch.test.ts
@@ -28,19 +28,19 @@ describe("Branch", () => {
     it("should be able to branch from hash", () => {
       const core = new GitgraphCore();
       const gitgraph = core.getUserApi();
+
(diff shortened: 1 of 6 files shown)


evaliyev:
@nicoespeon Please take a look

nicoespeon:
Sounds promising 👍 

I'd love we add tests on the core package & stories on the js/react ones to cover this, so we don't accidentally break it in the future. Could you add these?

If not, we'll do that with Fabien next Thursday.
Thanks for your contribution, really appreciated!

kohakukun:
@nicoespeon Thanks for the reply. I'll work on adding those tests and stories

kohakukun:
Added the requested changes.
@nicoespeon  PTAL

image

nicoespeon:
@kohakukun thanks a lot for your contribution! 🙏 
It has been released today in the different libs.

cc @evaliyev

PR #306, opened 2019-07-26
Pass graph as a prop to react component

evaliyev:
Good evening! 

Thanks for creating such a nice library! 

I have a use case where I have to keep updating the graph and update it on demand. This implies that I could create an instance of `GitgraphCore` and render it later. However, the React component does not allow such usage. That means I have to populate graph for the scratch every time I mount the component. 

It would be really nice to have a workflow where React component behaves only as a pure renderer and data to be rendered could be controlled from outside.

Kind regards,
Elchin.

### DIFF
--- a/packages/gitgraph-react/src/Gitgraph.tsx
+++ b/packages/gitgraph-react/src/Gitgraph.tsx
@@ -51,6 +51,7 @@ export {
 
 interface GitgraphProps {
   options?: GitgraphOptions;
+  graph?: GitgraphCore<ReactSvgElement>;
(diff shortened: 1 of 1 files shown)


evaliyev:
@nicoespeon @fabien0102 Please take a look

nicoespeon:
I just released this in gitgraph@react v1.5.0 👍 

Here's a Story to illustrate:

https://github.com/nicoespeon/gitgraph.js/blob/57cd512b0f170191f332eeda081e90feb005b8a1/packages/stories/src/gitgraph-react/1-basic-usage.stories.tsx#L21-L37

PR #315, opened 2019-07-27
Fix commit display without message

evaliyev:
This PR is created to show an existing bug in commit rending in `gitgraph/core`. It is most likely introduced by commit 5f022ef4b4458b9b851ff74cbed42588367801c0. 

@nicoespeon @fabien0102 Please take a look. I could try to fix it myself if you want. I just thought that you can fix it much faster. 

Cheers,
Elchin.

### DIFF
--- a/packages/gitgraph-core/src/__tests__/gitgraph.getRenderedData.style.test.ts
+++ b/packages/gitgraph-core/src/__tests__/gitgraph.getRenderedData.style.test.ts
@@ -1,7 +1,12 @@
 import { GitgraphCore } from "../gitgraph";
 import { Mode } from "../mode";
 import { BranchOptions } from "../branch";
-import { metroTemplate, TemplateName, blackArrowTemplate } from "../template";
+import {
+  metroTemplate,
+  TemplateName,
+  blackArrowTemplate,
+  templateExtend,
+} from "../template";
 import { Orientation } from "../orientation";
 
 describe("Gitgraph.getRenderedData.style", () => {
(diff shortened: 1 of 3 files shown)


nicoespeon:
@evaliyev Oh yes, the problem is certainly related to these lines:

https://github.com/nicoespeon/gitgraph.js/blob/2702d9275e5bd97cfe1230810ba20f380a811402/packages/gitgraph-core/src/user-api/branch-user-api.ts#L277-L283

The difference with previous behaviour is that `display` is now **always set**, not just set to `false`. And since it's declared after `style`, we override whatever configuration.

Thanks for creating the story. This specific behaviour was clearly not tested, or we'd have caught it before 👍 

What needs to be done:
- create a unit test to show the bug (instead of the story)
- fix it

I won't have time to do it today, but I'll try to fix it tomorrow, unless you have time to tackle it before.

evaliyev:
@nicoespeon Please take a look. I tried to make the fix as concise as possible. In the meantime, it could be confusing for others.

packages/gitgraph-core/src/user-api/branch-user-api.ts
```diff
@@ -280,8 +280,10 @@ class BranchUserApi<TNode> {
         ...withoutUndefinedKeys(
           this._branch.commitDefaultOptions.style!.message,
         ),
-        display: this._graph.shouldDisplayCommitMessage,
         ...style.message,
+        ...withoutUndefinedKeys({
+          display: this._graph.shouldDisplayCommitMessage && undefined,
```
nicoespeon: I see what you did there 😉 

I'm fine with that. I'll add a comment to explain so the intent is clear.

Possible alternatives would have been:
- using a temporary variable, as before (more verbose though)
- doing something like `...(!this._graph.shouldDisplayCommitMessage && { display: false })` but honestly, I think this would be obscure too 😁

packages/gitgraph-core/src/__tests__/gitgraph.getRenderedData.style.test.ts
```diff
@@ -197,6 +202,26 @@ describe("Gitgraph.getRenderedData.style", () => {
...
+
+    const core = new GitgraphCore({ template });
+    const gitgraph = core.getUserApi();
+
+    gitgraph.commit();
+
+    const { commits } = core.getRenderedData();
+    const [commit] = commits;
+
+    expect(commit.style.message.display).toBe(false);
```
nicoespeon: Awesome, thanks for adding the unit test 👍

nicoespeon:
@evaliyev I was about to give it a look. I was really happy to see you fixed the bug with a test and all 👏 

Gonna merge this one and I will release this as a fix.

evaliyev:
@nicoespeon Thanks a lot! It is a real pleasure to work with you :)

PR #320, opened 2019-08-13
Allow padding to be changed in template options (gitgraph-js)

tflori:
No styling option should be hardcoded. 

When the user decides the spacing of commits should be smaller then the labels are height, the labels and tags overlap each other and you can't decrease the size of the labels (at least not smaller than 22px).

I'm almost sure the same issue exists in the tags. But for now tell me what you think about before I invest more time in it.

### DIFF
--- a/packages/gitgraph-core/src/template.ts
+++ b/packages/gitgraph-core/src/template.ts
@@ -95,6 +95,14 @@ interface BranchLabelStyle {
    * Branch label border radius
    */
   borderRadius: number;
+  /**
+   * Branch label padding on x axis
+   */
+  paddingX: number;
+  /**
+   * Branch label padding on y axis
+   */
+  paddingY: number;
 }
 
 type BranchLabelStyleOptions = Partial<BranchLabelStyle>;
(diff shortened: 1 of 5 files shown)


nicoespeon:
Hi @tflori and thanks for your contribution.

You're right, it's preferable for styling options to be configurable.

The code looks correct, however there is more work that needs to be done. Automatic checks are failing because the option we're trying to reach doesn't exist. The core lib needs to change to reflect that, so everything pass.

image

Also, it would be nice to provide the same fix to the react lib, if you can.

Let's know if you need indication to implement that change completely. But the TS compilation should guide you through the required changes. When it compiles again, it should be good to go 👍

nicoespeon:
@tflori awesome, thanks for fixing CI.

One last thing: could you add a new story to illustrate the change? Let me know if you're not sure how to, we can guide you.

Thank you very much for your help!

tflori:
can I just add this style information to `"with custom branch labels"`?

tflori:
sorry I just saw that it is not working correct.. in gitgraph.ts the `BRANCH_LABEL_PADDING_X` is used to determine the size of the branch label. With this change you will get overlapping branch labels:
grafik

It is moving but not as expected. And to be honest: I know what I don't draw a graph by myself and using a library :slightly_smiling_face: 

IMO the style and how the padding is added to the label should only be known to the branch label function and this should return the boundaries of what was drawn. But what do I know... Sorry, enough pixel pushing for me - the rest is up to you.

nicoespeon:
Closing this since it won't probably be continued.

PR #330, opened 2019-10-18
Fix ineffective merge when fastForward is set to false

xmagpie:
Hello! Thanks a lot for this beautiful library!

Recently I stumbled on a problem with very slow performance during `merge`  without `fastForward` operation of `BranchUserApi` class. I have found out that despite `fastForward` argument being set to `false` an expensive `_areCommitsConnected` was still run. I've changed logic a bit to address the issue.


nicoespeon:
Hey @xmagpie 👋 

Awesome. All tests still pass, so everything is good. 

Thanks for improving the perfs of Gitgraph. You rock 👍

PR #340, opened 2019-12-16
Change Tag SVG radius value

nemonie:
One of the radius values of Tag is invalid. It throw an error in react environment.

```
Error: <path> attribute d: Expected number, ...
```


nicoespeon:
Thank you @nemonie, everything is green so let's merge this 👍

Issue #277, opened 2019-04-15
Gitgraph import JSON doesn't work with partial data


Andrey-Pavlov:
**Describe the bug**
If you have commits filtered by date range you can't render Git graph. The same if you have missed parent nodes.

**To Reproduce**
Remove any commit with id from any parentIds array.

**Expected behavior**
Expect to have "open" starts and ends.

**Screenshots**
I expect:
![My1](https://github.com/Andrey-Pavlov/GitGraphJSProblems/blob/master/4.15.2019/PartialData.png?raw=true)

**Additional context**
In my project (kinda CD/CI tool) I have the functionality to show delta commits between TeamCity builds. I want to show Git dag for this delta commits.

Symbolk:
Same problem for me, when I want to just show the latest N commits from git2json.

Symbolk:
@Andrey-Pavlov A workaround is to manually set the parents of the last commit to empty, to make the gitgraph renderable.

For example, display at most 20 commits:
```javascript
      git2json.run().then(gitlog => {
        // display at most 20 commits for performance
        if (gitlog.length <= 20) {
          this.gitGraph.import(gitlog)
        } else {
          let slicedLog = gitlog.slice(0, 20)
          slicedLog[slicedLog.length - 1].parents = []
          this.gitGraph.import(slicedLog)
        }
      })
```
@nicoespeon Do you have any better idea?

nicoespeon:
@Symbolk I don't, I think it's the simplest way to "cut" the graph today indeed 👍

Symbolk:
> @Symbolk I don't, I think it's the simplest way to "cut" the graph today indeed 👍

However, this workaround sometimes leads to an error:

![image](https://user-images.githubusercontent.com/14107297/63266004-89547100-c2c1-11e9-9f15-ef699b38260c.png)

PR #352, opened 2020-04-03
fix undefined current when importing parent branches

hlolli:
Because I want to visualize the commits from only last few days in a huge git repository, I hit an error on current version "cannot get parent of undefined". This should fix it. Never too safe with the nulls.


nicoespeon:
Nice!

@hlolli did you use the `import` feature with data of the last few days to reproduce this? If so, I think that might fix https://github.com/nicoespeon/gitgraph.js/issues/277 what do you think?

hlolli:
@nicoespeon yes exactly, we have a huge git repo at my company and we are hoping to visualize only last few days, it seems as any incomplete git history causes this. I had to patch git2json to be able to add that git option so I guess it can be hard to reproduce.

hlolli:
Looking at #277 I think it's the same error as I was getting.

PR #366, opened 2020-06-17
fix: replace 'transparent' with 'none'

Nebula83:
The code contains a couple of references to "transparent" as value for fill. When copying the svg data to file and opening with an SVG editor (for instance, inkscape, but every other converter I tried too) it causes misrendering by using the default black per SVG file standards. Probably because the keyword "transparent" is not supported. "none" seems to yield the same effect and is probably more in line with what you want to express. This PR fixes that.

References: 
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fill


nicoespeon:
@all-contributors please add Nebula83 for code

Issue #328, opened 2019-10-11
👋 Looking for maintainers


nicoespeon:
Hi everyone!

I just became dad few days ago 🎉 🍼  

Thus, it became clear to me that I need to focus my energy and **limit the number of project I run in parallel**. In terms of personal open-source side-projects, I'll focus my efforts on [Abracadabra](https://github.com/nicoespeon/abracadabra), a VS Code extension that provides automated refactorings for JS / TS.

It has been a long time I'm running Gitgraph.js with my friend @fabien0102. We created this project so we could easily draw simple git graphs, directly in the browser, to use for blog posts and presentations. These days however, **I don't use it much anymore**.

However, you can see with the currently open issues that people have different, interesting suggestions to cover more use-cases. Also, there are bugs to fix. Finally, the overall documentation of the library could be improved.

I won't have the time & energy to tackle these. Not in my foreseeable future at least. Thus, **I'm looking for new people to maintain the library and take over the project**.

I'll be able to provide guidance and onboard you on the project if you want. You'll see that the dev environment is "not that bad", with features essentially covered with tests (unit & E2E).

Let me know if you're interested. In the meantime, be aware that I won't work on the open issues myself.

![hack-tv](https://user-images.githubusercontent.com/1094774/66678520-1967c480-ec3a-11e9-91b2-11bfd6a84f02.gif)

marekyggdrasil:
@nicoespeon indeed managing such project is time consuming and it is understandable you have to prioritise to keep everything going.

I reviewed gitgraph.js repo and I find it understandable and test coverage is good as well.

I have some fresh ideas, I think this library could be used for more than just git graphs, personally I use it to take notes (in my research work I often have to explore many ideas simultaneously and at some point I realised git graphs with branches are good way to keep track of what is going on).

As you said you don't use use it much anymore, while for me it is a tool I use daily at work so I would like to help you out and dedicate some of my time to keep it running, resolve issues, bugs, help with docs etc.

If you want me to join in, feel free to reach me either here, e-mail or gitter, whatever the mean of communication you prefer.

nicoespeon:
@marekyggdrasil awesome. I'm on mobile right now but I'll reach you out next week to set you up with everything you need. Thank you, a lot ❤️

nicoespeon:
@marekyggdrasil I don't know if you saw that, but I replied to you with all details on Gitter 😉

marekyggdrasil:
[...]

PR #376, opened 2020-08-25
Updated computeGraphMap import

H3RSKO:
It was giving me a 'not found' error on this without change.

### DIFF
--- a/packages/gitgraph-node/src/index.ts
+++ b/packages/gitgraph-node/src/index.ts
@@ -1,7 +1,7 @@
 import { GitgraphCore } from "@gitgraph/core";
 
 // Domain (rendering logic)
-import computeGraphMap from "./compute-graph-map";
+import computeGraphMap from "./compute-graph-map.test.ts";
 
 // Infrastructure (logger implementations)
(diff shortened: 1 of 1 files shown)


nicoespeon:
Hey @H3RSKO 👋 

I think the problem should be somewhere else, the original import is supposed to import this file: https://github.com/nicoespeon/gitgraph.js/blob/master/packages/gitgraph-node/src/compute-graph-map/index.ts

The proposed change can't be valid because the source code is now importing a test file, which is unexpected.

Also, I remember this package wasn't quite ready to be used, so I wouldn't be surprised it's not working.

Finally, it seem CI is broken because versions are now incompatible and should be upgraded (that's not related to your PR, but it revealed that). It will be something to fix before accepting more PRs.

Unfortunately, I'm still looking for new maintainers for this project as I'm not working on it anymore: #328 

In the meantime, I'm closing this PR. If it still not work on your side (it might be an editor config or something), I suggest you point to `./compute-graph-map/index.ts` explicitly instead. But usually, this is the default behavior of node.

H3RSKO:
Got it. Thanks!

PR #385, opened 2020-09-25
Fix infrastractural problems

andreasonny83:
- Commit VSCode settings to enable Prettier on `ts` and `tsx` files
- Fix missing declarations in Storybook files that caused the develop script to fail
- Add `@gitgraph/core` as a direct dependency of `@gitgraph/stories`
- Fix `tsconfig` typo
- Add sleep to storybook script to fix async issues caused by `@gitgraph/core` not available


packages/stories/src/gitgraph-js/6-custom-renders.stories.tsx
```diff
@@ -25,7 +25,7 @@ const withoutBranchLabels = templateExtend(TemplateName.Metro, {
   branch: { label: { display: false } },
 });
 
-const renderSax: CommitOptions["renderDot"] = (commit) =>
+const renderSax: CommitOptions["renderDot"] = (commit: any) =>
```
nicoespeon, line 28: How come the type isn't inferred by TS here? Did you get failures?

packages/stories/package.json
```diff
@@ -17,6 +17,7 @@
     "build": "build-storybook"
   },
   "dependencies": {
+    "@gitgraph/core": "1.4.5",
```
nicoespeon, line 20: Oh you're right, we depend directly on the Core package. See https://github.com/nicoespeon/gitgraph.js/blob/884f8496fcf754217ca7e067820ed71b4b88693d/packages/stories/src/gitgraph-react/1-basic-usage.stories.tsx#L4

Good catch 👍

packages/stories/.storybook/tsconfig.json
```diff
@@ -1,22 +0,0 @@
-{
```
nicoespeon, line 1: Was it useless?

package.json
```diff
@@ -24,7 +24,7 @@
     "lerna": "lerna",
     "build": "lerna run build --scope=@gitgraph/{core,react,js,node} --stream",
     "watch": "lerna run watch --stream",
-    "storybook": "lerna run storybook --stream",
+    "storybook": "sleep 3 && lerna run storybook --stream",
```
nicoespeon, line 27: This may look mysterious in the future, but hopefully this PR description will bring the relevant context behind this 😉

nicoespeon (approved):

packages/stories/.storybook/tsconfig.json
andreasonny83: yes, there was no typescript files in that folder

package.json
andreasonny83: yep, that's actually a common problem when using concurrently and still need one task to be completed before the second one

Issue #389, opened 2020-11-06
React.StrictMode breaks the horizontal text alignments


ncortines:
Hi,

first of all sorry for the low effort bug report - I'll provide more details on request but I found the issue quite straightforward to reproduce.

When wrapping the react component with **React.StrictMode**, all the text horizontal alignments are broken.

Basically the translate(x) alignments are double compensated with a second translate(-x), leaving them all left aligned with the surrounding container.

Easy workaround is of course remove React.StrictMode form the application.

Thanks,
Juan

crutchcorn:
A minimal reproduction of this issue is here:

https://github.com/crutchcorn/graphjs-strict-bug

I'll be taking a look at this to see if I can't just make a PR right now

PR #391, opened 2020-11-21
[React] Individual file breakout

crutchcorn:
While investigating the bug #389, I looked in the developer console of React Devtools and found the following:

before

However, this is not a representative tree of what is being rendered in the `GitGraph.tsx` file. That's because instead of creating new components, we were simply using functions to render the DOM itself. I've broken out the functions into their own files/components, which makes debugging and code maintenance significantly easier:

after
 
There's still lots more work to do in terms of refactor effort, but I think this is a step in the right direction. I've tested this in my application and have no discovered any performance or behavior regressions.

I also plan on adding user testing to this library in the near future

### DIFF
--- a/packages/gitgraph-react/src/Commit.tsx
+++ b/packages/gitgraph-react/src/Commit.tsx
@@ -3,7 +3,7 @@ import {
   GitgraphCore,
   Commit,
   Mode,
-  Coordinate
+  Coordinate,
 } from "@gitgraph/core";
 import { CommitElement, ReactSvgElement } from "./types";
 import { Dot } from "./Dot";
@@ -24,6 +24,7 @@ interface CommitsProps {
   };
   getWithCommitOffset: (props: any) => Coordinate;
   setTooltip: (val: React.ReactElement<SVGGElement> | null) => void;
+  setCurrentCommitOver: (val: Commit<ReactSvgElement> | null) => void;
 }
 
 export class CommitComp extends React.Component<CommitsProps, {}> {
@@ -43,7 +44,7 @@ export class CommitComp extends React.Component<CommitsProps, {}> {
           <Tooltip commit={commit}>
             {commit.hashAbbrev} - {commit.subject}
           </Tooltip>
-        </g>
+        </g>,
       );
     }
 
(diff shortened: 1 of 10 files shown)


packages/gitgraph-react/src/Gitgraph.tsx
```diff
@@ -168,267 +166,32 @@ class Gitgraph extends React.Component<GitgraphProps, GitgraphState> {
...
-          coordinates.map((a) => a.map((b) => this.getWithCommitOffset(b))),
-          isBezier,
-          this.gitgraph.isVertical,
-        )}
-        fill="none"
-        stroke={branch.computedColor}
-        strokeWidth={branch.style.lineWidth}
-        transform={`translate(${offset}, ${offset})`}
-      />
+        <BranchPath
```
nicoespeon, line 183: 👍 Much better indeed

packages/gitgraph-react/src/Gitgraph.tsx
```diff
@@ -168,267 +166,32 @@ class Gitgraph extends React.Component<GitgraphProps, GitgraphState> {
...
-
-    if (!this.commitsElements[commit.hashAbbrev]) {
-      this.initCommitElements(commit);
-    }
-
-    this.commitsElements[commit.hashAbbrev].tags.push(ref);
-
-    return ref;
-  }
-
```
nicoespeon, line 431: 🧹 👏 Nice cleanup of the main file

packages/gitgraph-react/src/Commit.tsx
```diff
@@ -0,0 +1,135 @@
...
+  initCommitElements: (commit: Commit<ReactSvgElement>) => void;
+  commitsElements: {
+    [commitHash: string]: CommitElement;
+  };
+  getWithCommitOffset: (props: any) => Coordinate;
+  setTooltip: (val: React.ReactElement<SVGGElement> | null) => void;
+  setCurrentCommitOver: (val: Commit<ReactSvgElement> | null) => void;
+}
+
+export class CommitComp extends React.Component<CommitsProps, {}> {
```
nicoespeon, line 30: What is `Comp` referring to? Components?

What do you think about naming it `Commit` for consistency with the file name and other components?

nicoespeon (approved):
I like this change, I think it's a good thing you did that.

Also, we do have visual regression tests already. As far as CI can tell, you break none while doing this refactor. That's good enough for me 👍 

https://www.chromatic.com/build?appId=5c9632d9ead30500207bb576&number=181

packages/gitgraph-react/src/Commit.tsx
crutchcorn: Yup `Comp` is just short for "Component". I can totally rename it to `Commit`, I was just having some issues with auto-importing in my IDE (it was trying to import `Commit` from other internal packages), but I'll make that change shortly

PR #392, opened 2020-11-21
[React] Fix strict mode label placement

### LINKED ISSUE #367: Why my svg width cannot fit the container?
**Describe the bug**

Why my svg width cannot fit the container?

image

**To Reproduce**

[codesandbox](https://codesandbox.io/s/react-git-graph-5sbeq?file=/src/SimpleGraph.tsx)

**Screenshots**

image

### LINKED ISSUE #389: React.StrictMode breaks the horizontal text alignments

crutchcorn:
This PR builds on top of the work done in #391, it is expected to merge that PR in first.

Fixes #389 

Fixes #367

## Screenshots

Here's the broken version:

broken

Here is it fixed:

offsetfixed

## Why did this happen?

React StrictMode disallows and breaks unintended side-effects. The previous method of repositioning the labels was based on a kludgy but clever solution of keeping (homebrewed) `ref`s of every commit inside of a map. Once `componentDidUpdate` ran, it would run through the `ref`s and manually run a side-effect in order to calculate and update the element's `transform` property. This was done using Regex and the DOM APIs

However, that side effect was not very React-like and broke a lot of the timing behaviors, which is why it broke when `StrictMode` was enabled. As such, the solution I took was to assign React `useRef`s on every element during the `Commit` component render. Once `ref`s were assigned, I utilized the `useLayoutEffect` hook (which runs before browser paint, so we don't end up with nodes "jumping") to calculate their width, assign the x-offset values to `useState` hooks, and render those values alongside the actual values

This does a few things for us:

1) It fixes the bug
2) It eliminates "magic" positioning, which displaced the code for X positioning from the code for Y positioning. It is now much easier to follow the positioning code logic
3) Because it consolidates the positioning logic code, it makes these components much more "component-like" and makes further maintenance easier than the previous monolithic system

## Further

This PR is a massive one. I apologize for the breadth of changes in such a little period of time. I've done my best to bugtest, however I wouldn't be surprised if some sneaky buggers got in there. I appreciate the work of the development team to this point and hope this contribution doesn't come off negatively! Huge fan!

### DIFF
--- a/packages/gitgraph-react/src/Commit.tsx
+++ b/packages/gitgraph-react/src/Commit.tsx
@@ -28,8 +28,18 @@ interface CommitsProps {
 export const Commit = (props: CommitsProps) => {
   const {commit, commits, gitgraph, commitMessagesX} = props;
 
-  // This _should_ likely be an array, but is not in order to intentionally keep
-  // a bug in the codebase that existed prior to Hook-ifying this component
+  /**
+   * This _should_ likely be an array, but is not in order to intentionally keep
+   *  a potential bug in the codebase that existed prior to Hook-ifying this component
+   * @see https://github.com/nicoespeon/gitgraph.js/blob/be9cdf45c7f00970e68e1a4ba579ca7f5c672da4/packages/gitgraph-react/src/Gitgraph.tsx#L197
+   * (notice that it's a single `null` value instead of an array
+   *
+   * The potential bug in question is "what happens when there are more than one
+   * branch label rendered? Do they overlap or cause the message X position to be
+   * in the wrong position?"
+   *
+   * TODO: Investigate potential bug outlined above
+   */
   const branchLabelRef = React.useRef<SVGGElement>();
   const tagRefs: MutableRefObject<SVGGElement[]> = React.useRef([]);
   // "as unknown as any" needed to avoid `ref` mistypings later. :(
--- a/packages/gitgraph-react/src/Commit.tsx
+++ b/packages/gitgraph-react/src/Commit.tsx
@@ -10,7 +10,7 @@ import { Dot } from "./Dot";
 import { Tooltip } from "./Tooltip";
 import { Arrow } from "./Arrow";
 import { Message } from "./Message";
-import { Tag } from "./Tag";
+import { Tag, TAG_PADDING_X } from "./Tag";
 import { BranchLabel } from "./BranchLabel";
 import { MutableRefObject } from "react";
 
@@ -22,10 +22,11 @@ interface CommitsProps {
   getWithCommitOffset: (props: any) => Coordinate;
   setTooltip: (val: React.ReactElement<SVGGElement> | null) => void;
   setCurrentCommitOver: (val: Commit<ReactSvgElement> | null) => void;
+  commitMessagesX: number;
 }
 
 export const CommitComp = (props: CommitsProps) => {
-  const {commit, commits, gitgraph} = props;
+  const {commit, commits, gitgraph, commitMessagesX} = props;
 
   // This _should_ likely be an array, but is not in order to intentionally keep
   // a bug in the codebase that existed prior to Hook-ifying this component
@@ -33,6 +34,10 @@ export const CommitComp = (props: CommitsProps) => {
   const tagRefs: MutableRefObject<SVGGElement[]> = React.useRef([]);
   // "as unknown as any" needed to avoid `ref` mistypings later. :(
   const messageRef: MutableRefObject<SVGGElement> = React.useRef<SVGGElement>() as unknown as any;
+
+  const [branchLabelX, setBranchLabelX] = React.useState(0);
+  const [tagXs, setTagXs] = React.useState<number[]>([]);
+  const [messageX, setMessageX] = React.useState(0);
 
   const arrows = React.useMemo(() => {
     if (!gitgraph.template.arrow.size) return null;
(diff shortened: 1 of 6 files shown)


nicoespeon:
Hi @crutchcorn 👋 

Thanks for tackling that!

The PR description is very nice. I like the before/after examples and your explanation is very clear. I merged the other PR so first I think we can rebase this one before moving on.

Then, it's all good but we have a few regressions as you can see here: https://www.chromatic.com/build?appId=5c9632d9ead30500207bb576&number=182

3 scenarios don't look great after this change. There's likely something missing in the new logic that doesn't accommodate well with the "custom renderers" stories. Could you fix those?

image
image
image

nicoespeon:
@all-contributors please add crutchcorn for bugs, code

crutchcorn:
> I merged the other PR so first I think we can rebase this one before moving on.

I'll certainly rebase. 

I'm glad we have as strong of visual testing as we do - I was afraid of regressions (given the scope of work done) and wasn't sure how to test everything.

Luckily, I have a pretty strong idea of what's going on with custom renderers. Let me go ahead and try to fix those tonight (on top of rebasing this work) and see where we can go from there

crutchcorn:
I've now fixed the issue with the custom message rendered (turned out to be even simpler than I thought it was going to be!)

It's a good thing we revisited this PR as well - I went back to check against the storybook and discovered a bug that would throw an error on re-render (this has now been fixed and I did much more testing than I did originally - looks like everything works in both strictmode and not)

crutchcorn:
Discovered that this PR will also close #367. Two bugs for the price of one!

packages/gitgraph-react/src/Commit.tsx
```diff
@@ -1,135 +1,180 @@
 import * as React from "react";
 import {
   GitgraphCore,
-  Commit,
```
nicoespeon, line 4: Oh I see the conflict! Yep, 👍 for the alias

packages/gitgraph-react/src/Commit.tsx
```diff
@@ -1,135 +1,180 @@
...
-            />
-          }
-          {this.renderBranchLabels(commit)}
-          {this.renderTags(commit)}
-        </g>
-      </g>
-    );
-  }
+  // This _should_ likely be an array, but is not in order to intentionally keep
+  // a bug in the codebase that existed prior to Hook-ifying this component
```
nicoespeon, line 32: 👍 For the comment, it will be helpful for maintenance.

Could you add a word about the actual bug? Is there an open issue we could link here if people are looking for context?

packages/gitgraph-react/src/Gitgraph.tsx
```diff
@@ -112,21 +108,20 @@ class Gitgraph extends React.Component<GitgraphProps, GitgraphState> {
...
-                initCommitElements={this.initCommitElements.bind(this)}
-                commitsElements={this.commitsElements}
                 getWithCommitOffset={this.getWithCommitOffset.bind(this)}
                 setTooltip={this.setTooltip.bind(this)}
+                commitMessagesX={this.state.commitMessagesX}
               />
             )}
           </g>
-          {this.$tooltip}
+          {this.$tooltip}[
```
nicoespeon, line 124: ❗ This looks like a typo, nope?

```suggestion
          {this.$tooltip}
```

packages/gitgraph-react/src/Gitgraph.tsx
```diff
@@ -299,11 +243,3 @@ class Gitgraph extends React.Component<GitgraphProps, GitgraphState> {
     return { x, y: this.state.commitYWithOffsets[y] || y };
   }
 }
-
-function moveElement(target: Element, x: number): void {
```
nicoespeon, line 303: 🎉 Less hacky solution! Thank you!

packages/gitgraph-react/src/Message.tsx
```diff
@@ -32,7 +27,9 @@ export class Message extends React.Component<MessageProps> {
     const y = commit.style.dot.size;
 
     return (
-      <g ref={ref} transform={`translate(0, ${y})`}>
+      // 0, 14 on buggy
+      // 150, 14 on non-buggy
```
nicoespeon, line 31: 🧹  I think these were trails of debugging. I'd get rid of those now the bug is fixed.

packages/gitgraph-react/src/Tag.tsx
```diff
@@ -60,44 +60,18 @@ function DefaultTag(props: BaseTagProps) {
...
-    if (!this.props.commitsElements[commit.hashAbbrev]) {
-      this.props.initCommitElements(commit);
-    }
-
-    this.props.commitsElements[commit.hashAbbrev].tags.push(ref);
-
-    return ref;
-  }
-}
+export const Tag = React.forwardRef<SVGGElement, TagProps>((props, ref) => {
```
nicoespeon, line 66: Nice!

nicoespeon (approved):
Code looks good to me! I found a typo and some debug comments we can drop, but everything is fine for me!

That's even less code than before and indeed, easier to understand.

Also, CI is green so it means there's no regression in all the cases we support. That's a green light to me 👍 

@crutchcorn now that you're contributor, can you merge this PR? If so, I'll let you do so (you can since it's approved) after you've tackled the few comments.

crutchcorn:
Whoops! You're totally right about the typo and the debug comments. I'll axe those tonight and see about getting this merged in.

However, I can't seem to merge this PR. I don't appear to have permissions to do so

crutchcorn:
I've gone ahead and removed the typo and debug comments. Lastly, I've expanded what the "bug" (it's much more of a theoretical bug that I need to loop back around on later) is and what might cause it/etc. I think this should be good-to-merge now.

Just a reminder that I seemingly can't merge this PR myself, we might want to look into that.

PR #394, opened 2020-11-24
Fix linting glob

crutchcorn:
Originally, the linting glob was not only checking the `packages` folders, but all of the subfolders in `packages`. This included `node_modules` and as such, I was getting the following errors on every commit (due to linting being ran thanks to Husky)

```
​C:\Users\crutchcorn\git\OceanBit\graph\gitgraph.js [master ≡ +0 ~1 -0 !]
Î»  npm run lint

> gitgraph.js-monorepo@2.0.0-monorepo lint C:\Users\crutchcorn\git\OceanBit\graph\gitgraph.js
> tslint packages/**/src/**/*


ERROR: packages/gitgraph-js/node_modules/rollup-plugin-commonjs/src/ast-utils.js[50, 76]: == should be ===
ERROR: packages/gitgraph-js/node_modules/rollup-plugin-commonjs/src/index.js[50, 11]: Shadowed name: 'options'
ERROR: packages/gitgraph-js/node_modules/rollup-plugin-commonjs/src/is-cjs.js[10, 4]: The key 'promise' is not sorted alphabetically
ERROR: packages/gitgraph-js/node_modules/rollup-plugin-commonjs/src/transform.js[135, 29]: Shadowed name: 'name'
ERROR: packages/gitgraph-js/node_modules/rollup-plugin-commonjs/src/transform.js[160, 44]: Shadowed name: 'name'
ERROR: packages/gitgraph-js/node_modules/rollup-plugin-commonjs/src/transform.js[277, 13]: Shadowed name: 'name'
ERROR: packages/gitgraph-js/node_modules/rollup-plugin-commonjs/src/transform.js[298, 11]: Shadowed name: 'required'
ERROR: packages/gitgraph-js/node_modules/rollup-plugin-commonjs/src/transform.js[308, 10]: Shadowed name: 'required'
ERROR: packages/gitgraph-js/node_modules/rollup-plugin-commonjs/src/transform.js[377, 14]: Shadowed name: 'name'
ERROR: packages/gitgraph-js/node_modules/rollup-plugin-commonjs/src/transform.js[391, 4]: The key 'name' is not sorted alphabetically
ERROR: packages/gitgraph-js/node_modules/rollup-plugin-commonjs/src/transform.js[409, 4]: The key 'name' is not sorted alphabetically
ERROR: packages/gitgraph-js/node_modules/rollup-plugin-commonjs/src/transform.js[440, 12]: Shadowed name: 'name'
[...]

PR #395, opened 2020-11-24
Migrate linting to eslint

crutchcorn:
As mentioned at the end of #394, [TSLint has been depreciated for a year now](https://medium.com/palantir/tslint-in-2019-1a144c2317a9). This PR replaces the existing TSLint + Prettier in favor of ESLint + Prettier + TypeScript ESLint

This PR contains exactly 0 code logic changes, only the auto-format applied

While there are `0 warnings`, there _are_ 39 warnings.

With the exception of two "is defined but never used" warnings, the other 37 instances of warnings come from "Unexpected any" types. However, when investigating the usage of `any` in our codebase, I noticed that this would potentially introduce breaking changes to the APIs and I would like to break that work out into a separate PR (or more than likely, more than one PR)


package.json
```diff
@@ -39,16 +39,20 @@
...
+    "eslint-config-prettier": "^6.15.0",
+    "eslint-plugin-prettier": "^3.1.4",
+    "eslint-plugin-react": "^7.21.5",
     "husky": "1.1.2",
     "jest": "23.3.0",
     "lerna": "3.13.1",
-    "prettier": "1.13.7",
-    "pretty-quick": "1.6.0",
+    "prettier": "^2.2.0",
+    "pretty-quick": "^3.1.0",
```
nicoespeon, line 54: FYI, we used to prefer fixed versions to avoid bad surprises. This is especially true since it was a side-projects and we may not touch the dependencies for years. It prevents the "you come back 2 months later and nothing works" because someone didn't follow SemVer.

Do you have a preference for caret versions? If so, I'd be curious to discuss. 

☝ That's not blocking though.

package.json
```diff
@@ -28,7 +28,7 @@
     "develop": "concurrently npm:watch npm:storybook",
     "website": "lerna run start --scope=@gitgraph/website --stream",
     "deploy:website": "lerna run deploy --scope=@gitgraph/website --stream",
-    "lint": "tslint packages/**/src/**/*",
+    "lint": "eslint packages/*/src/**/*.{ts,tsx,js,jsx}",
```
nicoespeon, line 31: Of course this conflicts with master, but you'll be able to solve that easily 😉

packages/gitgraph-core/src/__tests__/gitgraph.clear.test.ts
```diff
@@ -37,15 +37,9 @@ describe("Gitgraph.clear", () => {
     const core = new GitgraphCore();
     const gitgraph = core.getUserApi();
 
-    gitgraph
-      .branch("dev")
-      .commit("one")
-      .commit("two");
+    gitgraph.branch("dev").commit("one").commit("two");
```
nicoespeon, line 40: Aha, new Prettier version 😆

nicoespeon (approved):
Scanned through the diff and that looks good to me 👍

packages/gitgraph-core/src/__tests__/gitgraph.clear.test.ts
crutchcorn: Haha yeah. I noticed that after working through this PR, but it's the default unfortunately

package.json
crutchcorn: > It prevents the "you come back 2 months later and nothing works" because someone didn't follow SemVer.

I can definitely relate to this on some level, but admittedly I've not experienced this problem in major ways in the past years. For reference, I maintain somewhat sizable libraries that have low-maintenance to them today:

- [PlopJS](https://github.com/plopjs/plop)
- [Node Vibrant](https://github.com/Vibrant-Colors/node-vibrant)

And have had a helping hand in even larger projects that have little maintenance to them nowadays and I've found that the offset of "hey, you need to bump your patch version to fix this audit issue" balanced out the few minor times that it takes to update a broken dep.

Even with libraries with extremely low-maintenance mode (like Node Vibrant, that I've effectively halted development for), I'm still more than willing to jump in if it's something as simple as locking a package bump for

That said, it's far from a hill I'm willing to die on - if you'd rather I switch to fixed versions I'm definitely up for it

nicoespeon: Ah that's fun, I was a contributor on PlopJS too: https://github.com/plopjs/plop/graphs/contributors

I see your point. I'm totally fine with it too and I think it's better you use the convention you're familiar with if you start maintaining it (again, I'm not so 🤷‍♂️ ) 👍

PR #399, opened 2021-03-05
Branch Deletion

### LINKED ISSUE #388: Deleting branches does not work
**Describe the bug**

`branch.delete()` does nothing

**To Reproduce**

Code for https://codepen.io/nicoespeon/pen/arqPWb?editors=0010

```
const graphContainer = document.getElementById("graph-container");
 
// Instantiate the graph.
const gitgraph = GitgraphJS.createGitgraph(graphContainer);

// Simulate git commands with Gitgraph API.
const master = gitgraph.branch("master");
master.commit("Initial commit");

var feature = master.branch("feature");
feature.commit("a");
master.merge(feature);
feature.delete();
```
Branch "feature" is still visible.

**Expected behavior**

The ref "feature" should be gone.

**Screenshots**

n/a

**Context**

`@gitgraph/js`

acransac:
This patch adds the possibility to delete a branch. It extends the branch user API so one can call `someBranch.delete()` and see the branch disappear from the rendered graph if it is not merged before.

Concretely, it clears out the branch name from the graph's `branches` property and from its `refs`. Commits and tags from the branch remain in the graph. Once a branch is deleted, the branch user API throws for the operations that don't make sense anymore. It is also impossible to delete a branch when it is checked out.

On the rendering side, only commits that are not reachable somehow disappear from view. This means that if the deleted branch was merged beforehand, the commits stay visible.

To the best of my understanding of the code, this change is compatible with previous use and behavior (notably, all tests pass without update). I also added tests and stories to cover what I added.

Closes #388

### DIFF
--- a/packages/gitgraph-core/src/__tests__/branch.test.ts
+++ b/packages/gitgraph-core/src/__tests__/branch.test.ts
@@ -188,4 +188,126 @@ describe("Branch", () => {
       });
     });
   });
+
+  describe("delete", () => {
+    let gitgraph;
+    let develop;
+    let feature;
+
+    beforeEach(() => {
+      gitgraph = new GitgraphCore().getUserApi();
+
+      develop = gitgraph.branch("develop");
+
+      develop.commit("develop first");
+
+      feature = gitgraph.branch("feature");
+
+      feature.commit("feature first");
+    });
+
+    it("should be deleted and not referenced", () => {
+      develop.checkout();
+
+      feature.delete();
+
+      expect(feature._branch.isDeleted() && !feature._isReferenced()).toBe(
+        true,
+      );
+    });
+
+    it("should leave commits and tags from the deleted branch in the graph", () => {
+      feature.tag("some tag");
+
+      const featureCommit = gitgraph._graph.refs.getCommit("feature");
+
+      develop.checkout();
+
+      feature.delete();
+
+      expect(
+        gitgraph._graph.commits.find(({ hash }) => hash === featureCommit) &&
+          gitgraph._graph.refs.hasCommit(featureCommit) &&
+          gitgraph._graph.tags.hasName("some tag"),
+      ).toBe(true);
+    });
+
+    it("should throw if the branch is checked out", () => {
+      expect(() => feature.delete()).toThrow(
+        `Cannot delete the checked out branch "feature"`,
+      );
+    });
+
+    it("should throw when branching from a deleted branch", () => {
+      develop.checkout();
+
+      feature.delete();
+
+      expect(() => feature.branch("other-feature")).toThrow(
+        `Cannot branch from the deleted branch "feature"`,
+      );
+    });
+
+    it("should throw when committing on a deleted branch", () => {
+      develop.checkout();
+
+      feature.delete();
+
+      expect(() => feature.commit("other commit")).toThrow(
+        `Cannot commit on the deleted branch "feature"`,
+      );
+    });
+
+    it("should silently do nothing when deleting twice", () => {
+      develop.checkout();
+
+      feature.delete();
+
+      feature.delete();
+
+      expect(feature._branch.isDeleted() && !feature._isReferenced()).toBe(
+        true,
+      );
+    });
+
+    it("should throw when merging a deleted branch to an existing one", () => {
+      develop.checkout();
+
+      feature.delete();
+
+      expect(() => develop.merge(feature)).toThrow(
+        `The branch called "feature" is unknown`,
+      );
+    });
+
+    it("should throw when merging some branch to a deleted branch", () => {
+      develop.checkout();
+
+      feature.delete();
+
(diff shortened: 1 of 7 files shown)


packages/stories/src/gitgraph-js/1-basic-usage.stories.tsx
```diff
@@ -389,4 +389,37 @@ storiesOf("gitgraph-js/1. Basic usage", module)
...
+        const master = gitgraph.branch("master").commit("Initial commit");
+        const develop = gitgraph.branch("develop");
+        develop.commit("one");
+        master.commit("two");
+        master.checkout();
+        develop.delete();
+      }}
+    </GraphContainer>
+  ))
+  .add("delete a branch after merging it", () => (
```
nicoespeon, line 409: 👍 Perfect, thanks for adding stories to this!

packages/gitgraph-core/src/user-api/branch-user-api.ts
```diff
@@ -119,6 +185,10 @@ class BranchUserApi<TNode> {
    */
   public merge(options: GitgraphMergeOptions<TNode>): this;
   public merge(...args: any[]): this {
+    if (this._branch.isDeleted() && !this._isReferenced()) {
+      throw new Error(`Cannot merge to the deleted branch "${this.name}"`);
```
nicoespeon, line 189: 👌 That makes sense and is a useful error to debug.

packages/gitgraph-core/src/user-api/branch-user-api.ts
```diff
@@ -98,6 +106,64 @@ class BranchUserApi<TNode> {
...
+   */
+  public delete(): this {
+    // Delete all references to the branch from the graph (graph.branches and graph.refs)
+    // and from the commits (commit.refs). Then, make the branch instance a deleted branch.
+    // Like in git, the commits and tags in the deleted branch remain in the graph
+    if (
+      this._graph.refs.getCommit("HEAD") ===
+      this._graph.refs.getCommit(this.name)
+    ) {
+      throw new Error(`Cannot delete the checked out branch "${this.name}"`);
```
nicoespeon, line 120: 👍 Nice, you thought about this edge-case

packages/gitgraph-core/src/__tests__/gitgraph.getRenderedData.branches.test.ts
```diff
@@ -221,4 +221,54 @@ describe("Gitgraph.getRenderedData.branches", () => {
       ]);
     });
   });
+
+  describe("deleted branch", () => {
+    let gitgraph;
```
nicoespeon, line 226: 🍰 We usually prefer to declare variables closer to their usage, directly into each test. It makes it easier to ensure each test is decoupled from the other (no shared variable).

It's OK here 'cause there are only 2 tests, but if you can refactor this it will be great!

Note that if we want to avoid repeating the same code statements to init the graph state, we can still extract them in a function instead, like so:

```ts
    it("should ignore commits on a deleted branch", () => {
      const { gitgraph, develop } = createGraphWithSomeCommits();

      develop.delete();

      const { commits } = gitgraph._graph.getRenderedData();

      expect(commits).toMatchObject([
        { subject: "one", refs: [] },
        { subject: "two", refs: ["master", "HEAD"] },
      ]);
    });
```

packages/gitgraph-core/src/__tests__/branch.test.ts
```diff
@@ -188,4 +188,126 @@ describe("Branch", () => {
       });
     });
   });
+
+  describe("delete", () => {
+    let gitgraph;
```
nicoespeon, line 193: 🍰 We usually prefer to declare variables closer to their usage, directly into each test. It makes it easier to ensure each test is decoupled from the other (no shared variable).

Note that if we want to avoid repeating the same code statements to init the graph state, we can still extract them in a function instead, like so:

```ts
    it("should be deleted and not referenced", () => {
      const { develop, feature } = createGraphWithSomeCommits();

      develop.checkout();

      feature.delete();

      expect(feature._branch.isDeleted() && !feature._isReferenced()).toBe(
        true,
      );
    });
```

nicoespeon (approved):
This PR is 💯  

- It has a lot of tests to prove that it's working—honestly, it also makes it easier to review as I can start reading the tests
- It has useful comments to explain "why" we do this
- It doesn't break other tests & stories, which means no existing behavior was changed

So 👏  I'm gonna ship this with the next release.

nicoespeon:
@all-contributors please add acransac for code, bug

packages/gitgraph-core/src/__tests__/branch.test.ts
acransac: Hi @nicoespeon, thanks for the quick review and kind feedback! Should I make a PR to refactor these tests?

